PR #854, opened 2018-11-16
Improve ctrl+C behavior around tools/apps

wildmolasses:
Hi Jay, I'm starting off a PR in response to #737. So far I made 2 quick edits. I'm going to search around the codebase and identify other places to fix. We can use this pull request to discuss, organize, and correct the changes I make.

Let's keep this pull request open until all desired ctrl+C changes are made.

### DIFF
--- a/src/tools/getBlock/getBlock.cpp
+++ b/src/tools/getBlock/getBlock.cpp
@@ -16,7 +16,7 @@
 //------------------------------------------------------------
 int main(int argc, const char * argv[]) {
 
-    etherlib_init();
+    etherlib_init(quickQuitHandler);
 
     // We want to get the latestBlock prior to turning on --prove for example
     COptions options;
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 local_config
 
 *.xcodeproj
+.idea
+.vscode
 objs
 build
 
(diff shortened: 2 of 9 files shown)


wildmolasses:
Notes while in tools folder:

I noticed in getAccounts.cpp that `acctlib_init()` does not accept a `QUITHANDLER`. Perhaps I should make an issue to add one.

ethName.cpp makes a call to `editFile()`. Therefore I didn't change the quithandler.

ethslurp.cpp writes to files in `Slurp()`so I didn't change the quithandler.

qblocks.cpp: quithandler remains default

tjayrush:
I commented on the issue. (Not sure if that was right. Should I comment on the PR?)

src/tools/getBlock/getBlock.cpp
```diff
@@ -16,7 +16,7 @@
 //------------------------------------------------------------
 int main(int argc, const char * argv[]) {
 
-    etherlib_init();
+    etherlib_init(quickQuitHandler);
```
wildmolasses: This quitHandler should be specified at a more granular level

tjayrush: Let's make the default behavior `quickQuitHandler` so end users will not have to specify this without getting the weird behavior of three-control+c to quit.

The easiest way to do this is like this:
1. Change the name of `defaultQuitHandler` to something more meaningful.
2. Leave `quickQuitHandler` as is
3. Remove the 'default' parameter to the function signature in the header and rebuild
4. The build will fail everywhere that doesn't have a handler specified
5. For all failures use the renamed (previously defaultHandler)
6. Get the build to work
8. Reenable default param in header
9. Remove use of quickQuit since it will now default
10. Rename quickQuit if you think it's worth it.

Issue #857, opened 2018-11-16
Printing a double rounded to a consistent number of decimals


tjayrush:
[This function](https://github.com/Great-Hill-Corporation/quickBlocks/blob/master/src/libs/utillib/conversions.cpp#L252) has been a perpetual pain. I can't figure out how to get this to print with

1. The decimal rounded to a certain number of decimal places including negative
2. The width of the string that gets printed is that same number of decimals, but takes up a consistent width (maybe a missing third param with default value). I need this because I frequently print tables of doubles and the get misaligned

```
What I want                                     What I get
     0.00000   0.00001       0.00000   0.00001
    10.00000   0.00001      10.00000   0.00001
  1000.00000   0.00001      1000.000   0.00001
  -120.00000   0.00001      -120.00000   0.00001
     0.00000   0.00001      0.00000   0.00001
```

It gets totally bad when I use tabs instead of spaces to separate the columns (I do this because I want to print to the screen and then possibly later import into Excel.

This is probably way easier than I'm making it out to be. There's a lot of formatting options on `cout` and `cerr` that I don't know how to use.

Also, create a test case under ./src/libs/test-libs/utillib by duplicating an existing test case.

wildmolasses:
Hi Jay, the hyperlink you posted in this issue is broken. Do you mind posting it?

tjayrush:
Fixed the comment above. https://github.com/Great-Hill-Corporation/quickBlocks/blob/master/src/libs/utillib/conversions.cpp#L252

tjayrush:
Closing

PR #871, opened 2018-11-19
Replace double_2_Str and add tests

wildmolasses:
Re: #857, fix `double_2_Str` and add tests.


src/libs/utillib/conversions.cpp
```diff
@@ -258,14 +261,10 @@ namespace qblocks {
...
-        memset(r, '\0', 100);
-        sprintf(s, "%.*g", (int)nDecimals, ((int64_t)(  pow(10, nDecimals) * (  fabs(f) - labs( (int64_t) f )  )  + 0.5)) / pow(10,nDecimals));  // NOLINT
-        if (strchr(s, 'e'))
-            s[1] = '\0';
-        sprintf(r, "%d%s", static_cast<int>(f), s+1);  // NOLINT
-        return string_q(r);
+        stringstream stream;
+        stream << fixed << setprecision((int)nDecimals) << f;
+        std::string output = stream.str();
+        return output;
```
tjayrush: I would usually just do this:

    return stream.str();

more terse plus doesn't create a temporary string that it immediately deleted.

src/libs/test-libs/utillib/convert/options.cpp
```diff
@@ -14,7 +14,7 @@
 
 //---------------------------------------------------------------------------------------------------
 static COption params[] = {
-    COption("~mode", "a number between 0 and 3 inclusive"),
+    COption("~mode", "a number between 0 and 4 inclusive"),
```
tjayrush: Love it

src/libs/test-libs/utillib/convert/testing.cpp
```diff
@@ -168,6 +168,18 @@ TEST_F(CThisTest, TestConverts_3) {
...
 
+//------------------------------------------------------------------------
+TEST_F(CThisTest, TestConverts_4) {
+        cout << "Running " << testName << "\n";
+
+        ASSERT_EQ("double_2_Str1", double_2_Str(14.15019, 5), "14.15019")
+        ASSERT_EQ("double_2_Str2", double_2_Str(14.15019, 3), "14.150")
+        ASSERT_EQ("double_2_Str3", double_2_Str(10586816.10315913591, 10), "10586816.1031591352") // should be 10586816.1031591359???
+        ASSERT_EQ("double_2_Str4", double_2_Str(10586816.10315913591, 2), "10586816.10")
+        ASSERT_EQ("double_2_Str5", double_2_Str(1050, 3), "1050.000")
```
tjayrush: Please add negative numbers and zero for both the float and the number of digits.

src/libs/utillib/conversions.cpp
```diff
@@ -12,6 +12,9 @@
  *-------------------------------------------------------------------------------------------*/
 #include "biglib.h"
 #include "conversions.h"
+#include <iomanip> // setprecision
+#include <sstream> // stringstream
```
tjayrush: Just a coding style thing. I usually put the system headers in a file called ./src/libs/utillib/basetypes.h (these two may already be included there).

If I do put system headers in a .cpp file, I put them first.

The reason is because the linter will tell me to do that when (and if) we run the linter.

src/libs/test-libs/utillib/convert/testing.cpp
tjayrush: Now that I think about it, this may be why I put that shirbird code in there to begin with. We need (and cannot do without) 18 decimal place accuracy. This is a hard stop.

In fact, the more I think about it, this is exactly why that original code was there to begin with. This, I think, is why Ethereum doesn't support floating point numbers.

I think this test is related https://github.com/Great-Hill-Corporation/quickBlocks/blob/287e7b3e624592c1a276044b65f7a92e849e08ff/src/libs/test-libs/etherlib/printFloat/printFloat.cpp#L32

wildmolasses: OK great. I will make sure we have accuracy. I am reading up on this stuff. Interestingly, both the old code and the new code fail the same tests. I'm digging into why.

wildmolasses:
Hi Jay, I'm still warming up to testing in qblocks. Can you tell me if the following workflow is correct?
- pull down latest changes, install (cmake ../src, make)
- run make test in build folder. This updates the test/working folder.
- Compare changes in test/working with test/gold.

If that is incorrect, can you advise? Perhaps I need to use a tool from #875

tjayrush:
It's complicated (like everything). Can you get on a call?

src/libs/test-libs/utillib/convert/testing.cpp
```diff
@@ -168,6 +168,36 @@ TEST_F(CThisTest, TestConverts_3) {
...
+
+        //ASSERT_EQ("double_2_Str13", double_2_Str(10586816.10315913591, 11), "10586816.10315913591")
+
+        // Significant figures
+        ASSERT_EQ("double_2_Str14", double_2_Str(1.100000000005), "1.1")
+        ASSERT_EQ("double_2_Str15", double_2_Str(0.), "0")
+        ASSERT_EQ("double_2_Str15", double_2_Str(0.000000000), "0")
+
+        ASSERT_EQ("addr_2_Str1", addr_2_Str("0x281055afc982d96fab65b3a49cac8b878184cb16"), "0x281055afc982d96fab65b3a49cac8b878184cb16")
+
```
tjayrush: Let's add two more test cases:
    double_2_Str(-1e-10, 4) == "-0.0001"
    double_2_Str(128,-1) == "120"

wildmolasses: @tjayrush the test cases you mention about should be `ASSERT_NOT_EQ`. Is that what you had in mind?

tjayrush: I made a mistake.  -1e4,5 should be -0.0001 And I just wanted to see what would happen with a negative input for width.

PR #969, opened 2019-01-03
Make rpc endpoint a configuration option

### LINKED ISSUE #538: feature request: add ability to make RPC calls to Parity nodes in addition to infura
Currently "remote" is hard coded to call infura. I would like to be able to set "remote:http://myhost:8545" to make RPC calls to arbitrary nodes.

### LINKED ISSUE #624: INFURA mode doesn't seem to work any more
Infrua twitter recently tweeted that they will be dropping support for pmainnet.infura.io which is what we use. Perhaps that's the problem.

### LINKED ISSUE #700: We need to formalize a ‘provider’ interface
Is there any way to set custom node URL or change Infura API key for EthSlurp util?

**User**
I am using EthSlurp, but looks like it stuck on run last times. I think this problem is connected with Infura restrictions. How can I change Infura node to my custom node or change Infura API key?
_[Note: The user should have said EtherScan everywhere in the above]_

**Me:**
EthSlurp doesn’t use Infura, it uses EtherScan. Do you have an API key for Infura or EtherScan? If EtherScan, then you need to go to EtherScan to get an API key. Here: https://etherscan.io/myapikey Please let me know if that works by posting to this issue.

**User**
I have read this issue before, so I thought that it uses Infura + EtherScan. I know how to receive EtherScan API key but how can I set my custom API key in the ethSlurp configuration?

**Me:**
There's a link here to get an API key or EtherScan https://etherscan.io/apis

wildmolasses:
Starts to resolve #624 
Resolves #538 
Sort of resolves #700 

I will open an issue to set the endpoint as an environment variable.

### DIFF
--- a/src/libs/etherlib/node_curl.cpp
+++ b/src/libs/etherlib/node_curl.cpp
@@ -24,7 +24,7 @@ namespace qblocks {
     CCurlContext::CCurlContext(void) {
         CToml toml(configPath("quickBlocks.toml"));
         headers      = "Content-Type: application/json\n";
-        baseURL      = toml.getConfigStr("settings", "rpcEndpoint", "http://localhost:8545");
+        baseURL      = getGlobalConfig()->getConfigStr("settings", "rpcProvider", "http://localhost:8545");
         callBackFunc = writeCallback;
         curlNoteFunc = NULL;
         theID        = 1;
--- a/src/libs/etherlib/node_curl.cpp
+++ b/src/libs/etherlib/node_curl.cpp
@@ -22,7 +22,6 @@ namespace qblocks {
 
     //-------------------------------------------------------------------------
     CCurlContext::CCurlContext(void) {
-        CToml toml(configPath("quickBlocks.toml"));
         headers      = "Content-Type: application/json\n";
         baseURL      = getGlobalConfig()->getConfigStr("settings", "rpcProvider", "http://localhost:8545");
         callBackFunc = writeCallback;
--- a/src/libs/etherlib/node_curl.cpp
+++ b/src/libs/etherlib/node_curl.cpp
@@ -22,8 +22,9 @@ namespace qblocks {
 
     //-------------------------------------------------------------------------
     CCurlContext::CCurlContext(void) {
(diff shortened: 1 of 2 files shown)


src/libs/etherlib/node_curl.cpp
```diff
@@ -22,8 +22,9 @@ namespace qblocks {
 
     //-------------------------------------------------------------------------
     CCurlContext::CCurlContext(void) {
+        CToml toml(configPath("quickBlocks.toml"));
         headers      = "Content-Type: application/json\n";
-        baseURL      = "http://localhost:8545";
+        baseURL      = toml.getConfigStr("settings", "rpcEndpoint", "http://localhost:8545");
         callBackFunc = writeCallback;
```
tjayrush: Two things:

1) We can use the global pointer to the configuration file which is loaded at the start of every program instead of opening the file again. You can remove the CToml line and replace the getConfigStr call with

    getGlobalConfig()->getConfigStr("settings", "rpcEndpoint", "http://localhost:8545")

2) I like the word 'rpcEndpoint' but let's use the word 'provider' as they do in web3 (here: https://web3js.readthedocs.io/en/1.0/web3.html#setprovider).  So same code but use 'provider' instead of 'rpcEndpoint'.

src/other/install/quickBlocks.toml
```diff
@@ -2,4 +2,5 @@
 current=0.2.0
 
 [settings]
-blockCachePath=<NOT_SET>
+rpcEndpoint=http://localhost:8545
```
tjayrush: Use 'provider' instead to mimic web3.js

src/libs/etherlib/node_curl.cpp
```diff
@@ -22,8 +22,9 @@ namespace qblocks {
 
     //-------------------------------------------------------------------------
     CCurlContext::CCurlContext(void) {
+        CToml toml(configPath("quickBlocks.toml"));
```
tjayrush: Remove this line

wildmolasses: Done

Issue #47, opened 2017-05-06
Autogenerated code with the same function name will overwrite other class definition files


tjayrush:
In monitors especially if two smart contracts have similar function names.

https://github.com/Great-Hill-Corporation/ethslurp/issues/124

tjayrush:
Closing for now. Will reopen once 0.2.0 is finished. Referenced by #293.

tjayrush:
There is a #define in grabABI.cpp that turns on an experiment that solves this issue. I think it works, but I've not checked that it works. Turn it on, and check that it works.

tjayrush:
This is a duplicate of #54

PR #1098, opened 2019-04-25
utillib: add git info to version line (branch, commit hash)

### LINKED ISSUE #1097: add git commit hash to console output
Add the latest commit hash such that...

```
Powered by QBlocks (GHC-TrueBlocks 0.5.5-alpha) [a40af3a]
```

wildmolasses:
Resolves #1097

### DIFF
--- a/src/libs/utillib/options_base.cpp
+++ b/src/libs/utillib/options_base.cpp
@@ -405,7 +405,7 @@ namespace qblocks {
         os << notes();
         if (!isReadme) {
             os << bBlue << "  Powered by QBlocks";
-            os << (isTestMode() ? "" : " (" + getVersionStr(true) + ")") << " [" << GIT_BRANCH << "@" << GIT_COMMIT_HASH << "]\n" << cOff;
+            os << (isTestMode() ? "" : " (" + getVersionStr(true) + ")") << "\n" << cOff;
         }
         string_q ret = os.str().c_str();
         return postProcess("usage", ret);
--- a/src/libs/utillib/options_base.cpp
+++ b/src/libs/utillib/options_base.cpp
@@ -405,7 +405,7 @@ namespace qblocks {
         os << notes();
         if (!isReadme) {
             os << bBlue << "  Powered by QBlocks";
-            os << (isTestMode() ? "" : " (" + getVersionStr(true) + ")") << "\n" << cOff;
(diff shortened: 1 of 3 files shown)


src/libs/utillib/options_base.cpp
```diff
@@ -405,7 +405,7 @@ namespace qblocks {
         os << notes();
         if (!isReadme) {
             os << bBlue << "  Powered by QBlocks";
-            os << (isTestMode() ? "" : " (" + getVersionStr(true) + ")") << "\n" << cOff;
+            os << (isTestMode() ? "" : " (" + getVersionStr(true) + ")") << " [" << GIT_BRANCH << "@" << GIT_COMMIT_HASH << "]\n" << cOff;
```
tjayrush: This is great, but do me a favor? Can you (1) enter the command `getBlock --latest`. You should see the Parity version string:

```
Parity-Ethereum//v2.3.4-beta-0e95db11d-20190220/x86_64-macos/rustc1.32.0
```

Let's make our string look like this up to but NOT including the operating system. So,

```
GHC-TrueBlocks//v2.3.4-alpha-COMMIT-YYYYMMDD
```

C++ has built-in __YEAR__ defines I think.

Also, did you think about putting this inside of getVersionStr? In that way, the change would take place everywhere. I think that would be okay.

tjayrush: Done. #47.

wildmolasses: I might like to include the branch name too, but I'll defer. The branch name can sort of be inferred by the commit hash.

wildmolasses: Do you want the date the user built the software here, or the date of the commit?

tjayrush:
Hmmmm…

Commit, I think.

If we ever produce compiled binaries, I would think that a user who builds the same code should get identical version strings.

-------------------------------
Thomas Jay Rush
TrueBlocks.io
jrush@trueblocks.io
(610) 937-2593

wildmolasses:
Ok, should we go off UTC timestamp then?

tjayrush:
Always UTC. As if there were nothing else. Useless complication otherwise. Thanks.

-------------------------------
Thomas Jay Rush
TrueBlocks.io
jrush@trueblocks.io
(610) 937-2593

wildmolasses:
Included YYYYMMDD in version string. Not 100% sure I've got the timestamp locale correct.

PR #1127, opened 2019-05-05
blockScrape: refactor and possibly rephrase

wildmolasses:
Hi Jay, I'm opening this PR so we can talk about blockScrape -- hopefully we can comment directly on lines of code here. This PR won't be "ready for merge" for a while. The intention is to use it as a discussion zone.

The first thing I did was rename handle_freshen to handle_scrape. This change, as well as all changes I make, should be considered suggestions more than anything.

### DIFF
--- a/src/apps/blockScrape/scraper_context.cpp
+++ b/src/apps/blockScrape/scraper_context.cpp
@@ -258,8 +258,8 @@ void CScraper::finalizeIndexChunk(void) {
     string_q asciiFile = indexFolder_sorted_v2 + padNum9(first)+"-"+padNum9(last) + ".txt";
     string_q binFile = indexFolder_finalized_v2 + padNum9(first)+"-"+padNum9(last) + ".bin";
 
-    CStringArray apps;
-    apps.reserve(options->maxIndexRows + 100);
+    CStringArray appearances;
+    appearances.reserve(options->maxIndexRows + 100);
 
     cerr << "Processing index.";
     cerr.flush();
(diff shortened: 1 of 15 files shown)


src/apps/blockScrape/scraper_context.cpp
```diff
@@ -258,38 +258,38 @@ void CScraper::finalizeIndexChunk(void) {
...
     size_t cnt = 0;
     ostringstream os;
-    for (auto app : apps) {
+    for (auto app : appearances) {
         os << app << "\n";
         if (!(++cnt % (options->maxIndexRows / 10))) {
             cerr << "."; cerr.flush();
         }
     }
     // TODO(jayrush): should this be stringToAsciiFile not append?
```
wildmolasses: makes sense. I scanned the file and, other than the declaration, asciiFile is not mentioned or touched -- except here in line 291.

wildmolasses:
It is in 9bf546707ccf1bc614da6c554fcb7010415a8608 that my pr's atomicity begins to unravel...

tjayrush:
I cannot even begin to tell you how happy I am to see this. Welcome! Please feel very free to make comments on anything you find confusing, interesting, less than optimal, wrong, etc.

tjayrush:
> It is in [9bf5467](https://github.com/Great-Hill-Corporation/quickBlocks/commit/9bf546707ccf1bc614da6c554fcb7010415a8608) that my pr's atomicity begins to unravel...

One concern I had was that if we changed these, we would be conflicting with existing names already named without _v2, but I searched the entire source tree and this is not a problem. Here's all the names:

```
indexFolder_v2
indexFolder_binary_v2
indexFolder_blooms_v2
indexFolder_finalized_v2
indexFolder_fixed_v2
indexFolder_fromwild_v2
indexFolder_pending_v2
indexFolder_sorted_v2
indexFolder_staging_v2
indexFolder_zips_v2
```

Some of these names a junk from temporary code (such as `fromwild` and `fixed`. Others may be slightly mis-named or confused such as 'pending', 'staging', etc. But this is a good start.

PR #1179, opened 2019-07-15
Merging in changes from docker branch

tjayrush:


tjayrush:
A re-write of the block scraper. The previous version worked, but the code was very poor quality. Cleanup. Needs testing.

PR #1189, opened 2019-08-01
generated options round 0.1

wildmolasses:
Hi Jay, this PR is not meant to be merged. It's merely an easy way for you to see what the option generator is doing. Please provide your feedback!


wildmolasses:
> I think this can be merged if you do a couple of things:
> 
> 1. What are the various parameters?

Here's the template string I'm using:

```
`\tCOption2("${option.command}",\t${option.core_alias !== "" ? `"${option.core_alias}"` : null},\t"${option.input_type}",\t${option.core_required},\t${option.core_visible},\t"${option.description_core}")`
```

I'm not super fond of the nondescriptive arguments either. Wondering if we can include the parameter name too? e.g. `COption2(option: "trans_list", short: "t", ...)`

> 2. What is the null column for? Is that the 'one-char' alias? There are no one-chars?

The code is generated from the csv. There are no one-chars specified in the csv yet.

> 3. Can you please use four spaces instead of tabs at the start of the line?

Sure, will make this change before merging PR in trueblocks-docker.

> 4. Put a comma after the last item in the list (I always do this since C++ doesn't care and I don't have to treat the last line if I move things around).

Will do.

> 5. How do you order the items by how they appear in the csv file?

Right now, yes. But there's an ordering column in the csv, and we can order by that (when it gets filled out)

> 6. The last item should always be the 'tool description' (i.e. empty command).

Sounds like we need to fill out the ordering column and implement the sorting logic in the template parser.

> 7. I'm not in love with , <on/off> UNLESS that is how it should look. I made this syntax up. Is there a better way? This field describes two different things: data type and available options (enum). Any thoughts?

As it is in the csv, so it shall be.

tjayrush:
> Here's the template string I'm using:
> 
> ```
> `\tCOption2("${option.command}",\t${option.core_alias !== "" ? `"${option.core_alias}"` : null},\t"${option.input_type}",\t${option.core_required},\t${option.core_visible},\t"${option.description_core}")`
> ```
> 
> I'm not super fond of the nondescriptive arguments either. Wondering if we can include the parameter name too? e.g. `COption2(option: "trans_list", short: "t", ...)`

Replace the leading tab with four spaces. Remove remaining tabs and replace with single space. It's very unusual to have tabs inside of the constructor definition. Replace leading tab with spaces because if you don't, when I edit the file, it will change it anyway and cause a diff.

tjayrush:
> The code is generated from the csv. There are no one-chars specified in the csv yet. So everywhere the one-char appears as `null`.

Use an empty string ("") instead of `null`. `null` is a pointer in `C`. It makes me write unnecessary code.

tjayrush:
> > 1. The last item should always be the 'tool description' (i.e. empty command).
> 
> Sounds like we need to fill out the ordering column and implement the sorting logic in the template parser.

Don't worry about this just yet. I put fake 'COption' items for `to_file` and `fmt` in some places. I will remove them from the csv and this won't be an issue (I think the items in the file are already in the correct order.)

tjayrush:
> > 1. I'm not in love with , <on/off> UNLESS that is how it should look. I made this syntax up. Is there a better way? This field describes two different things: data type and available options (enum). Any thoughts?
> 
> As it is in the csv, so it shall be.

Do we have a separate `type` column? Example values could be `address_list`, `address`, `block_list`, `unsigned int`, `enum`, etc. But then we might need an `enum` column (which would be fine).

tjayrush:
Assume the existence of bit-wise enums OPT_REQUIRED, OPT_FLAG, OPT_VISIBLE (although I would use hidden). Then instead of three or four parameters put them in a single parameter with 
OPT_REQUIRED | OPT_FLAG | OPT_VISIBLE -- use hidden since it's less prevelant.

PR #1191, opened 2019-08-03
Option tinkering

wildmolasses:


src/apps/chifra/options.cpp
```diff
@@ -9,7 +9,7 @@
 //---------------------------------------------------------------------------------------------------
 static const COption params[] = {
 // BEG_CODE_OPTIONS
-    COption2("command", "", "", OPT_REQUIRED | OPT_ARG, "one of [ leech | scrape | daemon | list | export | balances | stats | ls | rm | accounts | config | slurp | quotes | data | blocks | trans | receipts | logs | traces | state | abi ]"),
+    COption2("command", "", "<string>", OPT_REQUIRED, "one of [ leech | scrape | daemon | list | export | balances | stats | ls | rm | accounts | config | slurp | quotes | data | blocks | trans | receipts | logs | traces | state | abi ]"),
```
wildmolasses: oops, string is wrong. should be enum.

tjayrush:
I don't think I need this. The code that generates the COption2 lines is working and already incorporated.

PR #1293, opened 2019-10-19
add node provider choice

nysxah:
adding QuikNode dedicated node service to compliment API gateway service option


tjayrush:
@nysxah Thanks. Let me know if I can be of any help in getting TrueBlocks to work with QuickNode. It always seemed to me to be a perfect match.

PR #1586, opened 2021-04-25
Some small docs clean-up

MattDodsonEnglish:
Three different things here:
* Adds link to source code in README footers. My plan is to use these so that readers on the Docs site can jump straight to the repo they want.
* Removes old linux build instructions. Actually, I think there's some other stuff in `src/other` that could be removed too.
* Fixes broken link in ethSlurp README


tjayrush:
Looks good to me.

PR #1617, opened 2021-05-20
Develop

MattDodsonEnglish:
A few different edits, but all trying to improve readability and trim sentences:
* removed "intro" headings
* Tried to front sub-clauses and prepositional phrases (e.g. "at the beginning, do this")
* Creating line breaks at sentences ends or at 80 chars--this makes it much easier to review future git diffs.
* Adds special admonition formatting
* Includes two extra commits fixing typos that I introduced


MattDodsonEnglish:
Also, since these intros are separate from the autogenerated `--help` texts, it might make sense to move them to trueblocks-docs. Just to speed up the editing process by removing the need for PRs and merges.

PR #1632, opened 2021-06-08
Don't read uninitialized memory when getting username

### LINKED ISSUE #1630: getUserName can return uninitialized memory
I noticed that the /status API endpoint will sometimes emit raw bytes, and tracked it down to getUserName:

https://github.com/TrueBlocks/trueblocks-core/blob/fe94cf03d46df15229bcfce3a7126bb62b953c6d/src/libs/utillib/sfos.cpp#L309-L317

It looks like the return value of `getlogin_r` should be checked to see if it is zero, and on non-zero handle the case and return some type of initialized memory.

cole-h:
Fixes https://github.com/TrueBlocks/trueblocks-core/issues/1630.

---

Also defines `LOGIN_NAME_MAX` if `LOGIN_NAME_MAX` is not defined -- not if `HOST_NAME_MAX` is not defined.

### DIFF
--- a/src/apps/cacheStatus/options.cpp
+++ b/src/apps/cacheStatus/options.cpp
@@ -222,7 +222,7 @@ void COptions::Init(void) {
     mode = "";
 
 #ifndef HOST_NAME_MAX
-#define HOST_NAME_MAX 64
+#define HOST_NAME_MAX 256
 #endif
 #ifndef LOGIN_NAME_MAX
 #define LOGIN_NAME_MAX 64
--- a/src/apps/cacheStatus/options.cpp
+++ b/src/apps/cacheStatus/options.cpp
@@ -223,16 +223,15 @@ void COptions::Init(void) {
 
 #ifndef HOST_NAME_MAX
 #define HOST_NAME_MAX 64
+#endif
+#ifndef LOGIN_NAME_MAX
 #define LOGIN_NAME_MAX 64
 #endif
     char hostname[HOST_NAME_MAX];
     gethostname(hostname, HOST_NAME_MAX);
-    char username[LOGIN_NAME_MAX];
-    getlogin_r(username, LOGIN_NAME_MAX);
-    if (isDockerMode()) {
-        memset(username, 0, LOGIN_NAME_MAX);
+    char username[LOGIN_NAME_MAX] = { 0 };
+    if (getlogin_r(username, LOGIN_NAME_MAX) != 0 || isDockerMode())
         strncpy(username, "nobody", 7);
-    }
 
     if (isTestMode()) {
         status.host = "--hostname-- (--username--)";
--- a/src/libs/utillib/sfos.cpp
+++ b/src/libs/utillib/sfos.cpp
@@ -303,16 +303,15 @@ bool isRunning(const string_q& progName) {
 
 #ifndef HOST_NAME_MAX
 #define HOST_NAME_MAX 64
+#endif
+#ifndef LOGIN_NAME_MAX
 #define LOGIN_NAME_MAX 64
 #endif
 //----------------------------------------------------------------------------
 string_q getUserName(void) {
-    char username[LOGIN_NAME_MAX];
-    getlogin_r(username, LOGIN_NAME_MAX);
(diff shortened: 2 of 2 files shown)


src/apps/cacheStatus/options.cpp
```diff
@@ -223,16 +223,17 @@ void COptions::Init(void) {
 
 #ifndef HOST_NAME_MAX
 #define HOST_NAME_MAX 64
```
tjayrush: This change is great. Thanks so much for this. Let's correct this mistake as well while we're at it. Host names can be 256 characters long (actually 253, but we can use 256). Can you make that change to the HOST_NAME_MAX variable?

src/apps/cacheStatus/options.cpp
```diff
@@ -223,16 +223,17 @@ void COptions::Init(void) {
...
 #endif
     char hostname[HOST_NAME_MAX];
     gethostname(hostname, HOST_NAME_MAX);
-    char username[LOGIN_NAME_MAX];
-    getlogin_r(username, LOGIN_NAME_MAX);
-    if (isDockerMode()) {
-        memset(username, 0, LOGIN_NAME_MAX);
+    char username[LOGIN_NAME_MAX] = { 0 };
+    if (getlogin_r(username, LOGIN_NAME_MAX) != 0)
+        strncpy(username, "unknown", 8);
```
tjayrush: Is "unknown" something people would expect to see? In my experience (admittedly a bit outdated) this value would have been "nobody". If we use "nobody" we can collapse this with the `isDockerMode` clause.

src/libs/utillib/sfos.cpp
```diff
@@ -303,16 +303,17 @@ bool isRunning(const string_q& progName) {
...
-    if (isDockerMode()) {
-        memset(username, 0, LOGIN_NAME_MAX);
+    char username[LOGIN_NAME_MAX] = { 0 };
+    if (getlogin_r(username, LOGIN_NAME_MAX) != 0)
+        strncpy(username, "unknown", 8);
+    if (isDockerMode())
         strncpy(username, "nobody", 7);
-    }
     return username;
 }
```
tjayrush, line 317: Same comments here as above.

src/apps/cacheStatus/options.cpp
cole-h: Works for me. Only reason I put `unknown` is because that's more accurate -- when I observed this issue, it was running under a real user, but in a systemd service, which I suppose interfered with `getlogin_r` due to the lack of a TTY.

src/apps/cacheStatus/options.cpp
```diff
@@ -222,17 +222,16 @@ void COptions::Init(void) {
     mode = "";
 
 #ifndef HOST_NAME_MAX
-#define HOST_NAME_MAX 64
+#define HOST_NAME_MAX 256
+#endif
+#ifndef LOGIN_NAME_MAX
 #define LOGIN_NAME_MAX 64
 #endif
     char hostname[HOST_NAME_MAX];
```
andir, line 230: Might want to do go with `char hostname[HOST_NAME_MAX+1] = { 0 };` here:
```
gethostname() returns the null-terminated hostname in the character array name, which has a length of len bytes.  If the null-terminated hostname is too large to fit, then the name is truncated, and no error is returned (but see NOTES below).  POSIX.1 says that if such truncation occurs, then it is unspecified whether the returned buffer includes a terminating null byte.
```

tjayrush: @andir - the +1 writes past the end of the buffer. if MAX = 1, buffer[MAX] is one character long. buffer[MAX+1] writes to position 2. Not good.

andir: Read my proposal again. I am suggesting making the variable bigger not assigning to the n+1'th element.

tjayrush: My bad.

tjayrush: I made that change in the develop branch. Thanks for suggestion.

PR #1658, opened 2021-07-25
Fix arch linux build

RainM:
I fix build for the project on ArchLinux (highly likely caused by mainline compiler with additional warnings). There are two changes:

- query_flatfile.cpp - quite straightforward. I think issue cause is quite simple - mix of malloc and delete.
- node.cpp - not sure about this patch. As far as I see, previous version of if doesn't work (we compare NOPOS != NOPOS). So, there are two possible fixes: remove if statement at all or try to understand previous fix logic (it looks like there was typo 'lastTime' instead of 'thisTime').


tjayrush:
LGTM. Thanks for submitting...

PR #1670, opened 2021-08-03
Always add -j to the compilation in readme

LefterisJP:
Reasoning is that if you don't make tries to use max cores which many
times may hurt the system, use up all the memory and start swapping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ git clone -b develop git@github.com:TrueBlocks/trueblocks-core.git
 cd trueblocks-core
 mkdir build && cd build
 cmake ../src
-make
+make -j2
 ```
 
 (You may use `make -j <ncores>` to parallelize the build. Replace `<ncores>` with the number of cores on your machine.)


tjayrush:
I've seen this drag my computer down to a crawl (on a Mac and on an Ubuntu). For that reason, I tried to put `make -j nprocs` into the instructions, but that doesn't always work.

Hard coding -j2 means on a machine with 12 cores (like mine), only 2 will be used. Although, I guess 2 is better than 1 and it won't drag the machine down.

Any thoughts?

LefterisJP:
> I've seen this drag my computer down to a crawl (on a Mac and on an Ubuntu). For that reason, I tried to put `make -j nprocs` into the instructions, but that doesn't always work.
> 
> Hard coding -j2 means on a machine with 12 cores (like mine), only 2 will be used. Although, I guess 2 is better than 1 and it won't drag the machine down.
> 
> Any thoughts?

> Hard coding -j2 means on a machine with 12 cores (like mine)

Still 2 cores is pretty fast for compilation. Better 2, than using all 12 cores which is what the current suggestion does and slows the machine to a crawl. We can even do -j1 to be sure it works even at machines with 1 core, but I don't think those exist anymore and if they do they probably can't run and index trueblocks and the chain?

I mean sure you can try to use nproc by doing something like `make -j"$(($(nproc --all) / 2))"` to use half the available cores.
That works in my machine.

But what if the user has a different shell, what if there is no nproc, what if the syntax is somehow different?

Imo KISS and merge this :smile:

MattDodsonEnglish:
On a fresh Linux install, `make -j2`  quit at 25% on my machine. I have no idea whether this actually relates to the `-j` flag, but it failed twice, and `make` worked fine. I pasted the error messages below.

I wonder if there's a reliable way to make recommendations about `make` based on machine type. It seems like it varies considerably. Indeed what @LefterisJP  said about running bare `make`:

>Reasoning is that if you don't make tries to use max cores which many times may hurt the system, use up all the memory and start swapping

describes exactly what happened on my machine when we specified the cores. Running `make` on its own never had any issues for me.

Pasting the error, it was the same both times:

```
make -j2
[  0%] Built target git_version
Consolidate compiler generated dependencies of target util
Consolidate compiler generated dependencies of target ether
[ 10%] Built target util
[ 19%] Built target ether
Consolidate compiler generated dependencies of target pin
Consolidate compiler generated dependencies of target acct
[ 21%] Built target pin
Consolidate compiler generated dependencies of target sqlite3
[ 25%] Built target acct
[ 25%] Built target sqlite3
/bin/sh: line 1: go: command not found
make[2]: *** [go-apps/blaze/CMakeFiles/blaze.dir/build.make:70: blaze] Error 127
make[1]: *** [CMakeFiles/Makefile2:3676: go-apps/blaze/CMakeFiles/blaze.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
/bin/sh: line 1: go: command not found
make[2]: *** [go-apps/flame/CMakeFiles/flame.dir/build.make:70: flame] Error 127
make[1]: *** [CMakeFiles/Makefile2:3728: go-apps/flame/CMakeFiles/flame.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

tjayrush:
Let's put something like `make <options>` and then in the line below put something like "Note: In the above, <options> are up to you. You may be able to speed up your build with `make -j <nprocs>` but we've had reports that this crashes some machines. Don't say we didn't warn you."

and leave it at that. I run `make -j 12` and it builds super fast. For you, it crashes your machine. There's no one correct answer.

MattDodsonEnglish:
Oh wait, I'm a fool. I hadn't installed go, so the build failed anyway. But for some reason `make -j2` failed at 25%, and `make` failed at 99%. 

The other part I wrote, about unpredictability, I think are still valid.

MattDodsonEnglish:
>  like make <options> and then in the line below put something like "Note: In the above, are up to you. You may be able to speed up your build with make -j <nprocs> 

I like this. 

By the way, I just did it with `make -j 4` and it was pretty fast. Very fast, actually: 2 mins. If the build fails somewhere in the process, does the install pick up where it left off?

LefterisJP:
> > like make  and then in the line below put something like "Note: In the above, are up to you. You may be able to speed up your build with make -j
> 
> I like this.
> 
> By the way, I just did it with `make -j 4` and it was pretty fast. Very fast, actually: 2 mins. If the build fails somewhere in the process, does the install pick up where it left off?

Depends on when the compilation stopped before. Cmake does keep a cache afaik

PR #1687, opened 2021-08-11
Fixes TrueBlocks/trueblocks-core#1627 Fixed GH action by using correc…

dszlachta:
…t ports for both SSH connection and forwarding

- Added workflow_dispatch to allow running action manually
- Changed port to match host port policy
- Fixed port for SSH to connect to
- Check if tunnel is up

@tjayrush After merging it, we have to make sure that SSH_COMMAND secret is valid, let's talk about it during a call


dszlachta:
Closing this as a PR to run tests remotely will follow

PR #1697, opened 2021-08-13
Move tests to external machine

dszlachta:
- SSH to the testing machine
- Build TrueBlocks at the specific commit (`github.sha`)
- Build and run tests
- Report the result back to Github


tjayrush:
I am going to go ahead and merge this even though it won't work because my secrets are wrong. I'll fix them though.

PR #1698, opened 2021-08-13
[Work in Progress] Support EIP-1559 London

### LINKED ISSUE #1576: Changes to RPC with London Hard fork
https://hackmd.io/@timbeiko/1559-json-rpc

dszlachta:
This is a general PR for the ongoing work.
Changes implemented so far:

- [x] `baseFeePerGas` in block
- [ ] tests for `baseFeePerGas`

Fixes #1576

### DIFF
--- a/src/libs/etherlib/block.cpp
+++ b/src/libs/etherlib/block.cpp
@@ -78,6 +78,9 @@ string_q CBlock::getValueByName(const string_q& fieldName) const {
             if (fieldName % "blockNumber") {
                 return uint_2_Str(blockNumber);
             }
+            if (fieldName % "baseFeePerGas") {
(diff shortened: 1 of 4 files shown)


tjayrush:
Quick note. You won't have to build tests for `baseFeePerGas` nor any of the other tests. If you run the existing tests, they will fail if you add a new field to a class.

Also -- you can test the specific tool without having to run all the tests. For example, in the case of this change, you can do `test-all.sh tools/getBlocks` to test changes to the block class.

Not every test that is going to fail is going to fail when you do that, but most of them will. There will be failed tests elsewhere that you can create with `make test-all`. If you use that visual merge tool (Araxsis Merge) it makes copying the updated tests from `working` to `gold` a lot easier.

PR #1739, opened 2021-09-05
Add examples to API and fix formatting

MattDodsonEnglish:
First, this PR fixes the heading issue that you mentioned. It should look normal now. I also cleaned up trailing whitespace.

Second, this PR also adds example calls. If you think these look alright, I'm going to declare API documentation tasks "DONE" for now.


tjayrush:
Yes. I think we can call API tasks done for now. 

--------
Thomas Jay Rush
***@***.***

PR #1747, opened 2021-09-09
README; simplifies install instructions

MattDodsonEnglish:
If this commit's added to the README, the docs and readme instructions will
be identical.

This version of the instructions is more streamlined.
Although there's less hand holding, the beginning of the instructions
links to the "Troubleshooting" section on the docs' page.

That troubleshooting section has explicit directions for what to do at each
step. So I think this PR lets us have the best of the both worlds--streamlined
README for experienced users, and a guided docs site for those who need a bit
more help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ and edit configuration files.
 If you need help with a step, see the [installation's troubleshooting section](https://trueblocks.io/docs/prologue/installing-trueblocks/#troubleshooting).
 
 0. Install dependencies
-    - &#9745; [Install Go, version 1.12x or later](https://golang.org/doc/install).
+    - &#9745; [Install the latest version of Go](https://golang.org/doc/install).
     - &#9745; Install the other dependencies with your command line: `build-essential` `git` `cmake` `python` `python-dev` `libcurl3-dev` `clang-format` `jq`.
 
 1. Compile from the codebase
@@ -51,21 +51,22 @@ If you need help with a step, see the [installation's troubleshooting section](h
     cmake ../src
     make
     ```
-    _(You may use `make -j nproc` to parallelize the build.)_
+    _(You may use `make -j <ncores>` to parallelize the build. <Ncores> represents the number of cores to devote to the job)_
 
-2. Add `trueblocks-core/bin` to your shell PATH.
+2. Add `trueblocks-core/bin` to your shell `PATH`.
 
-3. Find your TrueBlocks configuration directory. It should be in either `$XDG_DATA_HOME`, or:
+3. Find your TrueBlocks configuration directory. It should be in one of these places:
 
     * On linux at `~/.local/share/trueblocks`
     * On mac at `~/Library/Application Support/TrueBlocks`
+    * If you've configured it, wherever the location of `$XDG_DATA_HOME` is
 
 4. In the configuration directory, edit `trueblocks.toml` to add your RPC and API keys. It should look something like this:
 ```toml
 [settings]
 rpcProvider = "<url-to-rpc-endpoint>"
-etherscan_key = "<key>"
 ```
+
 5. Test a command!
 ```shell
 chifra blocks 12345
--- a/README.md
+++ b/README.md
@@ -12,9 +12,7 @@ Please see this [Migration](https://github.com/TrueBlocks/trueblocks-core/blob/d
 
 ## Table of Contents
   - [Introduction](#introduction)
-  - [Prerequisites](#prerequisites)
-  - [Compiling TrueBlocks](#compiling-trueblocks)
-  - [Adding ./bin to your $PATH](#adding-bin-to-your-path)
+  - [Quick Install](#quick-install)
   - [Introducing chifra](#introducing-chifra)
   - [Using chifra](#using-chifra)
   - [Building the Index of Appearances](#building-the-trueblocks-index-of-appearances)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,76 +30,50 @@ TrueBlocks is a collection of libraries, tools, and applications that improve ac
...
+Some users have had success using WSL─you're on your own!
 
-For all operating systems, please follow [these instructions](https://golang.org/doc/install) to install `golang` on your system.
+These instructions assume you can navigate directories with the command line
+and edit configuration files.
+If you need help with a step, see the [installation's troubleshooting section](https://trueblocks.io/docs/prologue/installing-trueblocks/#troubleshooting).
 
-**For Linux**:
+0. Install dependencies
+    - &#9745; [Install Go, version 1.12x or later](https://golang.org/doc/install).
```
tjayrush: Do you have any problem with just saying 'latest' go version? I always encourage using the latest of everything. Maybe "latest stable version" or something.

README.md
```diff
@@ -32,76 +30,50 @@ TrueBlocks is a collection of libraries, tools, and applications that improve ac
...
+    - &#9745; Install the other dependencies with your command line: `build-essential` `git` `cmake` `python` `python-dev` `libcurl3-dev` `clang-format` `jq`.
```
tjayrush, line 44: I feel like we need to be specific for Ubuntu and Mac separately. Otherwise won't we get people being confused. I see that the previous was Linux specific, but this seems to leave them out to dry.

README.md
```diff
@@ -32,76 +30,50 @@ TrueBlocks is a collection of libraries, tools, and applications that improve ac
...
-```
+1. Compile from the codebase
+    ```shell
+    git clone -b develop https://github.com/trueblocks/trueblocks-core
+    cd trueblocks-core
+    mkdir build && cd build
+    cmake ../src
+    make
+    ```
+    _(You may use `make -j nproc` to parallelize the build.)_
```
tjayrush: If they use this command directly, it will fail. Shouldn't you use <nproc> to indicate that the need to run a program called nproc to figure it out?

README.md
```diff
@@ -32,76 +30,50 @@ TrueBlocks is a collection of libraries, tools, and applications that improve ac
...
-```shell
-brew install cmake
-brew install git
-brew install clang-format
-brew install jq
-```
+2. Add `trueblocks-core/bin` to your shell PATH.
 
-We recommend that you run MacOS Big Sur or later for best results.
+3. Find your TrueBlocks configuration directory. It should be in either `$XDG_DATA_HOME`, or:
```
tjayrush: 99.9% of people won't have this set in their environment.

README.md
```diff
@@ -32,76 +30,50 @@ TrueBlocks is a collection of libraries, tools, and applications that improve ac
...
-git clone -b develop https://github.com/trueblocks/trueblocks-core
-cd trueblocks-core
-mkdir build && cd build
-cmake ../src
-make -j2
+4. In the configuration directory, edit `trueblocks.toml` to add your RPC and API keys. It should look something like this:
+```toml
+[settings]
+rpcProvider = "<url-to-rpc-endpoint>"
+etherscan_key = "<key>"
```
tjayrush: We need to comment that this is optional. Otherwise people will (as two people already have) comment on how weird it is for us to claim 100% fully local and in the installation instructions we require a EtherScan key.

tjayrush:
Waiting for your responses to my comments to merge.

README.md
MattDodsonEnglish: No problem, changed to latest.

README.md
MattDodsonEnglish: My thinking is that, this README is for people who have a good sense of how to install packages. But I want to catch everything, so I added the widest possible dependencies.

This list includes everything that someone on a mac needs, and a little more. Actually, it's more even than what's needed on Ubuntu, which I believe typically comes preinstalled with git.

 Also, on the troubleshooting, I added specific examples for installing packages with Ubuntu and Mac.
 https://trueblocks.io/docs/prologue/installing-trueblocks/#how-do-i-install-packages-on-the-command-line

README.md
MattDodsonEnglish: I've changed it to this:
     _(You may use `make -j <ncores>` to parallelize the build. <ncores> represents the number of cores to devote to the make job)_
     
I am going to add a section called "How many cores can I use for make?" And mention installing `nproc`,

README.md
MattDodsonEnglish: I've removed this, and added section in troubleshooting that quickly states the precise moment when `etherscan` would be needed, and mentions that TrueBlocks is working on removing this. https://trueblocks.io/docs/prologue/installing-trueblocks/#how-do-i-add-external-rpcs-and-etherscan

MattDodsonEnglish:
OK, I've updated the README and site docs for your feedback. The only thing I didn't change was adding OS-specific instructions, because the troubleshooting section has guided installs, and for the other reasons I mentioned here: https://github.com/TrueBlocks/trueblocks-core/pull/1747#discussion_r707269288

README should be cleaner and faster now. I also think we might be able to remove the README's troubleshooting section, but that's a discussion for another day.

README.md
tjayrush: Sorry to be pendantic about this, but I'd like to try to put this to bed once and for all.

I would call myself very familiar with installing packages and using the command line. If I saw this line (which basically says -- go install a bunch of stuff), I would be very annoyed. I want copy and paste instructions for my operating system -- but -- I'm willing to do just Mac and Linux. (We don't need to do Windows, nor any other Linux version other than Ubuntu -- if someone is using non-Ubuntu, they will figure it out).

I see your trouble shooting section, which is good, but perhaps we can at least link to that directly here with something like "For operating specific command, see here."

The thing I want to avoid is annoying people like me who know how to do thing but are uber-lazy and want copy and paste commands that work without any thought on my part.

README.md
tjayrush: Looks good.

README.md
tjayrush: Looks good. 

Change "The articulate option requires an EtherScan API key." to "The articulate option currently requires an EtherScan API key."

PR #1779, opened 2021-10-01
Small fixes to data model docs

MattDodsonEnglish:
Two small changes:

- removes directional language
- removes a junk character

I know there's an autogeneration flow. So I've tried to make sure that I made the edits in the right places.


tjayrush:
Thanks Matt.

PR #1794, opened 2021-10-05
Feature/1772 chifra range validation

dszlachta:
- Moved validate.go to separate package
- Added block range parser
- Added block identifier (number, hash, range, special, date) validation

Only one test failed.

### DIFF
--- a/src/apps/chifra/cmd/abis_run.go
+++ b/src/apps/chifra/cmd/abis_run.go
@@ -15,31 +15,32 @@ package cmd
 import (
 	"strings"
 
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/validate"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/utils"
 	"github.com/spf13/cobra"
 )
 
 func validateAbisArgs(cmd *cobra.Command, args []string) error {
 	if AbisOpts.classes {
-		return usage("the '{0}' option is not implemented", "--classes")
+		return validate.Usage("the '{0}' option is not implemented", "--classes")
 	}
 
 	if len(AbisOpts.sol) > 0 {
 		cleaned := "./" + strings.Replace(AbisOpts.sol, ".sol", "", 1) + ".sol"
 		if !utils.FileExists(cleaned) {
-			return usage("file not found at {0}", cleaned)
+			return validate.Usage("file not found at {0}", cleaned)
 		}
 		return nil
 	}
 
 	if len(AbisOpts.find) == 0 && !AbisOpts.known {
-		err := validateOneAddr(args)
+		err := validate.ValidateOneAddr(args)
 		if err != nil {
 			return err
 		}
 	}
 
-	err := validateGlobalFlags(cmd, args)
+	err := validate.ValidateGlobalFlags(RootOpts.file, RootOpts.fmt, cmd, args)
 	if err != nil {
 		return err
 	}
--- a/src/apps/chifra/cmd/abis_run.go
+++ b/src/apps/chifra/cmd/abis_run.go
@@ -40,7 +40,7 @@ func validateAbisArgs(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	err := validate.ValidateGlobalFlags(RootOpts.file, RootOpts.fmt, cmd, args)
+	err := validateGlobalFlags(cmd, args)
 	if err != nil {
 		return err
 	}
--- a/src/apps/chifra/cmd/blocks_run.go
+++ b/src/apps/chifra/cmd/blocks_run.go
@@ -21,7 +21,7 @@ import (
 )
 
 func validateBlocksArgs(cmd *cobra.Command, args []string) error {
-	err := validate.ValidateGlobalFlags(RootOpts.file, RootOpts.fmt, cmd, args)
+	err := validateGlobalFlags(cmd, args)
 	if err != nil {
 		return err
 	}
--- a/src/apps/chifra/cmd/blocks_run.go
+++ b/src/apps/chifra/cmd/blocks_run.go
@@ -21,11 +21,6 @@ import (
 )
 
 func validateBlocksArgs(cmd *cobra.Command, args []string) error {
-	err := validateGlobalFlags(cmd, args)
-	if err != nil {
-		return err
-	}
-
 	validationErr := validate.ValidateBlockIdentifiers(
 		args,
 		validate.ValidArgumentBlockHash|
@@ -45,6 +40,11 @@ func validateBlocksArgs(cmd *cobra.Command, args []string) error {
 		}
 
 		return validationErr
+	}
+
+	err := validateGlobalFlags(cmd, args)
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/src/apps/chifra/cmd/blocks_run.go
+++ b/src/apps/chifra/cmd/blocks_run.go
@@ -15,11 +15,12 @@ package cmd
 import (
 	"fmt"
 
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/validate"
 	"github.com/spf13/cobra"
 )
 
 func validateBlocksArgs(cmd *cobra.Command, args []string) error {
-	err := validateGlobalFlags(cmd, args)
+	err := validate.ValidateGlobalFlags(RootOpts.file, RootOpts.fmt, cmd, args)
 	if err != nil {
 		return err
 	}
--- a/src/apps/chifra/cmd/blocks_run.go
+++ b/src/apps/chifra/cmd/blocks_run.go
@@ -13,6 +13,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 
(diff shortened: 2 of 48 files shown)


test/gold/tools/getBlocks/getBlocks_combo_two_range.txt
```diff
@@ -1,35 +1,29 @@
 chifra blocks  1590000 4000001-4000002 1001001-1001100 1234567
-
```
tjayrush, line 2: This test case actually shows a shortcoming of the old block range parser. How hard would it be, given the new parser, to support multiple block ranges on a single command line? Just curious. If it's easy, we can do it, if it's at all hard, not a big deal.

src/apps/chifra/pkg/validate/blockvalidate.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func IsBlockNumber(str string) bool {
+	_, err := strconv.ParseUint(str, 10, 64)
+
+	return err == nil
+}
+
+func IsSpecialBlock(str string) bool {
+	// We have no way to tell if a string is valid special block
```
tjayrush: Actually, this is not true. We (TrueBlocks) names 10-20 blocks in this code: https://github.com/TrueBlocks/trueblocks-core/blob/master/src/libs/utillib/options_base.cpp#L999. There's only really three 'special' blocks according to Ethereum: 'first', 'latest', and 'pending.' The rest, we've been collecting and naming ourselves. They are accurate (in the sense that they do name actual blocks), but they are arbitrary and could be either expanded or shunken depending on if we think they're useful. 

They never really worked on any tool other than `chifra when`, so if we can somehow bring them into the Go code and make them available for all tools that support block ranges, that would be great.

Having mentioned tools supporting block ranges, I now realize that `chifra export` and `chifra list` should support the same (they don't). One user in our discord asked this exact question just today. I wish I could have been able to answer him with "use a block range" because that's exactly what he wanted.

src/apps/chifra/cmd/abis_run.go
```diff
@@ -15,31 +15,32 @@ package cmd
...
 	if len(AbisOpts.find) == 0 && !AbisOpts.known {
-		err := validateOneAddr(args)
+		err := validate.ValidateOneAddr(args)
 		if err != nil {
 			return err
 		}
 	}
 
-	err := validateGlobalFlags(cmd, args)
+	err := validate.ValidateGlobalFlags(RootOpts.file, RootOpts.fmt, cmd, args)
```
tjayrush: I think that ValidateGlobalFlags should not be part of the Validate package. It's pretty specific to the tools, but not really general like block ranges, addresses, or transaction ids. Also, in this way, we could remove these two additional items.

In other words, move this one validation (validateGlobalFlags) into the `cmds` package.

src/apps/chifra/cmd/blocks_run.go
```diff
@@ -13,17 +13,40 @@
...
 )
 
 func validateBlocksArgs(cmd *cobra.Command, args []string) error {
-	err := validateGlobalFlags(cmd, args)
+	err := validate.ValidateGlobalFlags(RootOpts.file, RootOpts.fmt, cmd, args)
 	if err != nil {
 		return err
 	}
 
+	validationErr := validate.ValidateBlockIdentifiers(
```
tjayrush, line 24: I want to do all non-global validations before doing the global validations, so move the in front of the above (before line 22).

src/apps/chifra/cmd/explore_run.go
```diff
@@ -13,11 +13,12 @@
 package cmd
 
 import (
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/validate"
 	"github.com/spf13/cobra"
 )
 
 func validateExploreArgs(cmd *cobra.Command, args []string) error {
-	err := validateGlobalFlags(cmd, args)
+	err := validate.ValidateGlobalFlags(RootOpts.file, RootOpts.fmt, cmd, args)
```
tjayrush: Same comment as above about making validateGlobalFlags part of the `cmds` package.

src/apps/chifra/cmd/when_run.go
```diff
@@ -22,10 +25,25 @@ func validateWhenArgs(cmd *cobra.Command, args []string) error {
 	// 		return errors.New(fmtError("You must provide either a date or a block number"))
 	// 	}
 	// }
-	err := validateGlobalFlags(cmd, args)
+	err := validate.ValidateGlobalFlags(RootOpts.file, RootOpts.fmt, cmd, args)
 	if err != nil {
 		return err
 	}
+
+	validationErr := validate.ValidateBlockIdentifiers(args, validate.ValidArgumentAll, 1)
```
tjayrush, line 29: Move this before the global validation (before line 21)

src/apps/chifra/pkg/validate/blockvalidate.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	return bRange.StartType == blockRange.BlockRangeDate
+}
+
+func IsRange(str string) (bool, error) {
+	bRange, err := blockRange.New(str)
+
+	if err == nil {
+		if bRange.Start.Special == "latest" {
```
tjayrush, line 75: Just a double check. Does `chifra blocks latest` work? It should. I think this is already a test case, but just double checking...

src/apps/chifra/pkg/validate/validate.go
```diff
@@ -85,28 +99,28 @@ func validateEnum(field, value, valid string) error {
...
 			return err
 		}
 	}
 	return nil
 }
 
-func validateGlobalFlags(cmd *cobra.Command, args []string) error {
-	if len(RootOpts.file) > 0 && !utils.FileExists(RootOpts.file) {
-		return usage("file {0} not found", RootOpts.file)
+func ValidateGlobalFlags(file string, fmt string, cmd *cobra.Command, args []string) error {
```
tjayrush: Maybe we can move this function to the `cmds` package so we don't have to send in these extra strings.

tjayrush:
This looks generally excellent. I left a few comments. I'll wait until you respond to merge.

test/gold/tools/getBlocks/getBlocks_combo_two_range.txt
dszlachta: For validation only, it's super easy. 

We use a function to validate command-line arguments. This function lets us set the maximum number of ranges that can be supplied to a command. So setting this number to -1 would disable the check.

But then, C++ code would still fail, right?

src/apps/chifra/pkg/validate/blockvalidate.go
dszlachta: Oh, sorry I missed it. I will port this list to go code and change this function

src/apps/chifra/cmd/abis_run.go
dszlachta: 👍

src/apps/chifra/cmd/blocks_run.go
dszlachta: 👍

src/apps/chifra/pkg/validate/blockvalidate.go
dszlachta: Yes it does work. This check is only for ranges. We check for specials earlier

test/gold/tools/getBlocks/getBlocks_combo_two_range.txt
tjayrush: Good point. I'll write an issue. Maybe you can put in a TODO comment on how easy it would be.

tjayrush:
Write this issue, but otherwise, I'm ready to merge when you are. https://github.com/TrueBlocks/trueblocks-core/issues/1798

src/apps/chifra/pkg/specialBlock/list.go
```diff
@@ -0,0 +1,57 @@
...
+	"devcon4":        6610517,
+	"uniswap":        6627917,
+	"constantinople": 7280000,
+	"devcon5":        8700401,
+	"mcdai":          8928158,
+	"istanbul":       9069000,
+	"muirglacier":    9200000,
+	"berlin":         12244000,
+	"london":         12965000,
+	"latest":         -1,
```
dszlachta, line 34: @tjayrush I have to set it to -1, because the default "zero value" for int (`0`) is already taken by `first` block. This is different than in C++ code, but I think it will make do for now. And when we move some logic to golang, we can change this line

test/gold/tools/whenBlock/whenBlock_invalid_alone_syntax_4.txt
```diff
@@ -1,4 +1,23 @@
 chifra when  0pppppp
+Error: The given value '0pppppp' is not a numeral or a special named block.
+Usage:
+  chifra when [flags] < block | date > [ block... | date... ]
 
-  The given value '0pppppp' is not a numeral or a special named block.
```
tjayrush, line 3: I guess I must have been wrong when I said that this would resolve to '0' block. Just as well, it actually seems better that it doesn't pass through anyway.

PR #1812, opened 2021-10-08
Bugfix/blockrange octals

dszlachta:
Added support for numbers with leading zeroes as block numbers.

Before, Go would take strings like "024" as octals and convert them to decimals (resulting in 20, not 24)

### DIFF
--- a/src/apps/chifra/pkg/blockRange/parser.go
+++ b/src/apps/chifra/pkg/blockRange/parser.go
@@ -47,6 +47,8 @@ package blockRange
 // 2021-10-03T10:30:59-1000:100
 
 import (
+	"strconv"
+
 	"github.com/alecthomas/participle/v2"
 	"github.com/alecthomas/participle/v2/lexer"
 )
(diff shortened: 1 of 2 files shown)


dszlachta:
All failed tests are failing for other PRs as well, I doubt the failures have something to do with the changes on this branch

tjayrush:
A couple of questions for you:

1) Are there existing integration test cases that would fail if we didn't fix this?
2) SHOULD `chifra` accept numbers with leading zeros as a decimal or is this a bug with `chifra`?
3) I think this is an example of where we wouldn't need a new integration test case. Do you agree? If not, can you add a new integration test case?

Since we're still alpha should we fix this in the correct way (getting the C++ code to accept numbers with leading zeros as octal) instead of 'breaking' the natural behavior of Go? No one has ever sent an octal number to `chifra` anyway.

dszlachta:
If you ask me, decimals with leading zeroes are weird and I would not expect any user to expected "0234" to be valid block number for chifra. 

I ran into this issue when I tried to parse  Unchained Index manifest "newBlockRange" field as BlockRange type. But now I think I might have gone a bit too far with this:
- as you said, no one ever tried to input 0NNNN as a number
- this problem exists only for manifest
- the ranges in manifest are really a different type (they can't have date, special names or hexes in them)

However, `chifra when` C++ code just drop the 0. Let's talk briefly about it today and either delete this PR or merge it

tjayrush:
Oh. That's really interesting...we do use a fixed-width, nine-digit, leading-zero block range for the chunk indexes. We do this so the files sort lexically in the directory listing. (Do you see how deep our reliance on direct access to the machine is? -- I didn't do this on purpose, but we avoid a sort of the directly listing in this way.)

The ranges in the manifest are a different type. I think you're right.

dszlachta:
I think we can just close it. There's nothing that should be merged

PR #1850, opened 2021-10-29
Feature/go manifest parsing

dszlachta:
Scope of this PR:
- calling UnchainedIndex smart contract to get manifest's CID
- fetching manifest from IPFS
- parsing manifest as JSON or TSV
- handling configuration files
- handling stderr logging
- analysing which blooms/chunks should be downloaded (not re-downloading already present files)
- fetching and validating chunks
- uncompressing chunks
- locking files
- handling `--list`, `--init` and `--all`
- listing chunks as either table or JSON


src/apps/chifra/go.mod
```diff
@@ -3,8 +3,11 @@ module github.com/TrueBlocks/trueblocks-core/src/apps/chifra
 go 1.16
 
 require (
-	github.com/alecthomas/participle/v2 v2.0.0-alpha7
+	github.com/BurntSushi/toml v0.4.1 // indirect
```
tjayrush: If, before you push your code, you `cd ../src/apps/chifra` and run `go mod tidy` these `indirect` dependencies go away and become direct dependencies. I'm not sure if it matters, but when I build sometimes this file changes in this way and then I get a `diff` that I don't expect.

We could make this part of a local commit action so it gets done automatically, I suppose, but I don't know how to do that.

src/apps/chifra/pkg/rpcClient/client.go
```diff
@@ -0,0 +1,28 @@
...
+func Get() ethclient.Client {
+	client, err := ethclient.Dial(config.ReadGlobal().Settings.RpcProvider)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	return *client
+}
+
+func HexToAddress(hex string) common.Address {
```
tjayrush, line 35: Just a question: why create a separate function here? Why not just call common.HexToAddress directly? I suppose it's so we can support different clients (i.e. multi-chain)? Is that why?

src/apps/chifra/pkg/rpcClient/client.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"log"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+// Setups a client instance and returns it
```
tjayrush: Go comments should be of the form:

// FunctionName verb description

If they are, then `go doc` builds the documentation you see online. In this case it might say something like

// Get sets up a client instance and returns it.

src/apps/chifra/pkg/blockRange/range_test.go
```diff
@@ -119,3 +122,34 @@ func TestHandleParserErrors(t *testing.T) {
...
+func TestBlockRange_UnmarshalJSON(t *testing.T) {
+	type SomeRecord struct {
+		Blocks BlockRange `json:"blocks"`
+	}
+
+	var record SomeRecord
+	source := []byte(`{"blocks":"000000000-10567003"}`)
+
+	err := json.Unmarshal(source, &record)
+
```
tjayrush: Just a personal preference, but I would like the `err != nil` code to follow on the next line after the `err` is created in every case. Remove this blank line, please.

src/apps/chifra/pkg/blockRange/range.go
```diff
@@ -61,6 +62,24 @@ type BlockRange struct {
...
 
+func (br *BlockRange) UnmarshalJSON(data []byte) error {
+	str, err := strconv.Unquote(string(data))
+
+	if err != nil {
+		return err
+	}
+
+	newBlock, err := New(str)
+
```
tjayrush: Just a personal preference, but I would like the err != nil code to follow on the next line after the err is created in every case. Remove this blank line, please. I won't comment any more.

In general err != nil always follow directly after the err's creation. with no blank line.

src/apps/chifra/pkg/blockRange/range.go
```diff
@@ -61,6 +62,24 @@ type BlockRange struct {
 	Modifier     Modifier
 }
 
+func (br *BlockRange) UnmarshalJSON(data []byte) error {
+	str, err := strconv.Unquote(string(data))
+
```
tjayrush: Just a personal preference, but I would like the `err != nil` code to follow on the next line after the `err` is created in every case. Remove this blank line, please.

src/apps/chifra/pkg/config/utils.go
```diff
@@ -0,0 +1,76 @@
...
+	confDir := osToPath[userOs]
+	if len(confDir) == 0 {
+		log.Fatalln("Unsupported operating system")
+	}
+
+	return path.Join(homeDir, confDir, fileName)
+}
+
+// Opens the file and returns new TOML Decoder
+func read(fileName string) *toml.Decoder {
```
tjayrush: If this function always only opens .toml files, can we rename this function to readToml?

src/apps/chifra/pkg/config/utils.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		log.Fatalln("Could not read configuration file:", filePath)
+	}
+
+	return toml.NewDecoder(file)
+}
+
+// Reads {configName}.toml from the config directory, parses it
+// and loads the values into target. Warning: configName should NOT
+// have .toml extension (it will be added automatically)
```
tjayrush: Instead of warning the caller in a comment, we should either

1) fail this function if .toml is present in the input, or
2) append .toml to the input if it's missing, but use it if it's present (my preference)

src/apps/chifra/pkg/config/types.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Global struct {
+	Version  globalVersion
+	Settings globalSettings
+}
+
+// BlockScrape config
+
+type blockScrapeRequires struct {
+	Trancing bool
```
tjayrush: Misspelling - should be Tracing

Also, you can add these fields as we will need them eventually:

	Tracing bool
	Parity bool
	Balance bool
	Archive bool

src/apps/chifra/pkg/config/configuration.go
```diff
@@ -0,0 +1,42 @@
+package config
+
+var DefaultIpfsGateway = "http://gateway.ipfs.io/ipfs/"
+
+// Functions in this package read and cache configuration files,
+// filling in default values
+
+var cachedGlobal Global
+
+func ReadGlobal() *Global {
```
tjayrush, line 24: I know I asked you to create the PR before it was ready, but let's comment every function as we write it.

// FunctionName does a thing.

If your function name matches the first word in the comment, `go doc` will treat it right.

src/apps/chifra/pkg/config/configuration.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	ReadTo(config, "trueBlocks")
+	cachedGlobal = *config
+	return config
+}
+
+var cachedBlockScrape BlockScrape
+
+func ReadBlockScrape() *BlockScrape {
```
tjayrush, line 43: Comment missing. I won't repeat myself, you can add a comment to all routines. Never more than a sentence unless especially complicated.

src/apps/chifra/pkg/config/configuration.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	config := &BlockScrape{
+		Dev: blockScrapeDev{
+			IpfsGateway: DefaultIpfsGateway,
+		},
+	}
+
+	ReadTo(config, "blockScrape")
+	cachedBlockScrape = *config
```
tjayrush, line 65: How do default values get set? If the user has no value set there should be a reasonable default. We should always make sure the code will run even if every configuration file is missing.

src/apps/chifra/pkg/file/lock.go
```diff
@@ -0,0 +1,41 @@
...
+
+func Lock(file *os.File) error {
+	// Lock configuration. We want both read (F_RDLCK) and
+	// write (F_WRLCK) locks, and by Start to be interpreted
+	// as the start of the file (Whence set to SEEK_SET).
+	//
+	// More: https://man7.org/linux/man-pages/man2/fcntl.2.html
+	lockConfig := &syscall.Flock_t{
+		Type:   syscall.F_RDLCK | syscall.F_WRLCK,
+		Whence: int16(io.SeekStart),
```
tjayrush, line 30: Just a question: why is there an `int16` here? That can't be the actual size of `Wence`, can it? Our files are way bigger than 2^16 bytes big.

src/apps/chifra/pkg/logger/logger.go
```diff
@@ -0,0 +1,34 @@
...
+var severityToLabel = map[severity]string{
+	Progress:   "PROG",
+	Info:       "INFO",
+	Warning:    "WARNG",
+	Error:      "ERROR",
+	ErrorFatal: "FATAL",
+}
+
+func Log(sev severity, a ...interface{}) {
+	fmt.Fprintf(os.Stderr, "<%s> : ", severityToLabel[sev])
```
tjayrush: We can mimic Erigon's log formatting here:

```
INFO[10-29|22:03:54.326] [1/18 Headers] Processed                 highest inserted=13515748 age=12s
```

But, if `IsTestMode() == true` then don't print the timestamp as that will break test cases.

src/apps/chifra/pkg/pinlib/manifest/range.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"strconv"
+	"strings"
+)
+
+type ManifestRange [2]uint64
+
+func (mr *ManifestRange) UnmarshalJSON(data []byte) error {
+	unquoted, err := strconv.Unquote(string(data))
+
```
tjayrush: remove line

src/apps/chifra/pkg/pinlib/manifest/range.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return err
+	}
+
+	newManifestRange, err := ManifestRangeFromString(unquoted)
+
```
tjayrush: remove line

src/apps/chifra/pkg/pinlib/manifest/range.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return err
+	}
+
+	*mr = newManifestRange
+
+	return nil
+}
+
+func ManifestRangeFromString(source string) (ManifestRange, error) {
```
tjayrush: In all my C++ code I always name a function like this a_2_b, so the thing you send in is being turned into the thing you're returning.

In this case, it would be StringToManifestRange. Otherwise, the code becomes a confusing blend of From's and To's and it ultimately you get one of each that does the same thing. Let's always make this type of function a_2_b.

Also, I use `x_2_y` so I can search for _2_ (to distinguish it from 'to' which appears all over the place) and find every conversation in the entire code base. So this could be `Str_2_ManifestRange` but that's just a possibility.

src/apps/chifra/pkg/pinlib/save.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"os"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/file"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/pinlib/manifest"
+)
+
+func SaveManifest(filepath string, manifest *manifest.Manifest) error {
+	targetFile, err := os.Create(filepath)
+
```
tjayrush: Remove line.

src/apps/chifra/pkg/pinlib/save.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return fmt.Errorf("creating file: %s", err)
+	}
+
```
tjayrush: Remove line.

src/apps/chifra/pkg/pinlib/save.go
```diff
@@ -0,0 +1,46 @@
...
+	defer targetFile.Close()
+
+	err = file.Lock(targetFile)
+
```
tjayrush: Remove line.

src/apps/chifra/pkg/pinlib/save.go
```diff
@@ -0,0 +1,46 @@
...
+			pin.BloomHash,
+			pin.IndexHash,
+		})
+	}
+
+	w := csv.NewWriter(targetFile)
+	w.Comma = '\t'
+	w.WriteAll(records)
+
+	if err := w.Error(); err != nil {
```
tjayrush, line 49: Can we move this up three lines to line 36? I'd like to try to always check errors immediately after they're created. In this way, as I can code, I can easily ignore it.

src/apps/chifra/pkg/pinlib/save.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"encoding/csv"
+	"fmt"
+	"os"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/file"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/pinlib/manifest"
+)
+
+func SaveManifest(filepath string, manifest *manifest.Manifest) error {
```
tjayrush, line 25: Needs a comment.

src/apps/chifra/pkg/pinlib/list.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	return fileNames, nil
+}
+
+func ListLocalBlooms() ([]string, error) {
+	return ListByChunkType(chunk.BloomChunk)
+}
+
+func ListLocalIndexes() ([]string, error) {
```
tjayrush, line 54: Comment

src/apps/chifra/pkg/pinlib/list.go
```diff
@@ -0,0 +1,38 @@
...
+		fileNames = append(
+			fileNames,
+			cacheLayout.RemoveExtension(file.Name()),
+		)
+	}
+
+	return fileNames, nil
+}
+
+func ListLocalBlooms() ([]string, error) {
```
tjayrush, line 48: Comment

src/apps/chifra/pkg/pinlib/list.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"io/ioutil"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/pinlib/chunk"
+)
+
+// This package makes it easier to get file names of all blooms/index chunks
+// that are currently saved locally
+
+func ListByChunkType(chunkType chunk.ChunkType) ([]string, error) {
```
tjayrush, line 26: Comment

src/apps/chifra/pkg/pinlib/manifest/json_manifest.go
```diff
@@ -0,0 +1,15 @@
+package manifest
+
+import (
+	"encoding/json"
+	"io"
+)
+
+func ReadJSONManifest(reader io.Reader) (*Manifest, error) {
```
tjayrush, line 21: Comment

src/apps/chifra/pkg/pinlib/download.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"errors"
+	"net/http"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/pinlib/manifest"
+)
+
+func DownloadManifest(gatewayUrl string) (*manifest.Manifest, error) {
+	response, err := http.Get(gatewayUrl)
+
```
tjayrush: Remove line

src/apps/chifra/pkg/pinlib/contract.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"os"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/rpcClient"
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+)
+
+var unchainedIndexAddr = "0xcfd7f3b24f3551741f922fd8c4381aa4e00fc8fd"
```
tjayrush: I think both of these items may want to be in a config file.

src/apps/chifra/pkg/pinlib/contract.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return "", fmt.Errorf("while parsing contract ABI: %w", err)
+	}
+
+	unpacked, err := contractAbi.Unpack("manifestHash", response)
+	if err != nil {
+		return "", fmt.Errorf("while unpacking value: %w", err)
+	}
+
+	return unpacked[0].(string), nil
```
tjayrush, line 80: Shouldn't we always check that `len(unpacked) > 0` first? It's not at all clear from above this `Unpack` always returns something if it doesn't err. What if the `contractABI` doesn't contain `manifestHash`? (It does, but what if it doesn't for some reason?

src/apps/chifra/pkg/pinlib/manifest/manifest.go
```diff
@@ -0,0 +1,36 @@
+package manifest
+
+import (
+	"os"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
+)
+
+type IpfsHash = string
```
tjayrush, line 15: Eventually, I think we may store this as []byte but I'm not sure why. In Erigon, they always prefer []byte to string for some reason? Size, speed? Not sure.

src/apps/chifra/pkg/pinlib/manifest/manifest.go
```diff
@@ -0,0 +1,36 @@
...
+	BloomHash IpfsHash `json:"bloomHash"`
+	IndexHash IpfsHash `json:"indexHash"`
+}
+
+type Manifest struct {
+	FileName           string          `json:"fileName"`
+	IndexFormat        IpfsHash        `json:"indexFormat"`
+	BloomFormat        IpfsHash        `json:"bloomFormat"`
+	CommitHash         string          `json:"commitHash"`
+	PreviousHash       IpfsHash        `json:"prevHash"`
```
tjayrush, line 28: This whole data structure is totally wrong in the C++ code. After we merge it, I will be changing this to make it much, much easier and more simple to understand.

It was intended in the old C++ code to make updating an existing unchained index cache on disc faster by being able to download only new files, but that happens already so this is all extraneous.

src/apps/chifra/pkg/pinlib/manifest/manifest.go
```diff
@@ -0,0 +1,36 @@
...
+	NewBlockRange      ManifestRange   `json:"newBlockRange"`
+	PreviousBlockRange ManifestRange   `json:"prevBlockRange"`
+	NewPins            []PinDescriptor `json:"newPins"`
+	PreviousPins       []PinDescriptor `json:"prevPins"`
+}
+
+func FromLocalFile() (*Manifest, error) {
```
tjayrush: Move this function to the `tab_manifest.go` file since it deals with the non-JSON manifest file format.

src/apps/chifra/pkg/pinlib/manifest/tab_manifest.go
```diff
@@ -0,0 +1,77 @@
...
+			BloomHash: record[1],
+			IndexHash: record[2],
+		})
+	}
+
+	return descriptors, nil
+}
+
+func BuildTabRange(descriptors []PinDescriptor) (ManifestRange, error) {
+	firstPinRange, err := ManifestRangeFromString(descriptors[0].FileName)
```
tjayrush: What if `len(descriptors) == 0`?

src/apps/chifra/pkg/pinlib/manifest/tab_manifest.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return ManifestRange{
+		firstPinRange[0],
+		lastPinRange[1],
+	}, nil
+}
+
+func ReadTabManifest(r io.Reader) (*Manifest, error) {
+	descriptors, err := ReadPinDescriptors(r)
+
```
tjayrush: Remove line

src/apps/chifra/pkg/pinlib/manifest/tab_manifest.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	newBlockRange, err := BuildTabRange(descriptors)
+
```
tjayrush: Remove line

src/apps/chifra/pkg/pinlib/manifest/tab_manifest_test.go
```diff
@@ -0,0 +1,40 @@
...
+013320355-013323254	QmYLcU9qUpLhETffC4X2MjtKkFZRDn85XXBgGTMT2sWhv1	QmZqZ2ZUZ3s8tHBm34dGWyXpr5ShZgkHpTKg1n79NpehSJ
+013317545-013320354	QmRxGboQht2ST1NMwkWdytsF1nL5FVrgfEhZ4ozBU1UZQU	QmQP9oKDZrqBDXMXCJS5UjWhtp8viDuAJqy3t2ERgFs6X7
+013314548-013317544	QmXFzVi2TE2iR2ALeJVfGe7ZFgeDgjW3UeHiAussgFvQoX	QmZbT4zLcp8WTM6MH33D1nVP4eRCDwGTk2L37rud2c1bgK
+013311678-013314547	QmaBmWyq4KdiuMrrErxgniduejueqQi497DzVfvzdR3Ft8	QmP72KErK2Z6rbpQ8tHjTtTqUrMEEimaBtk9vzxunKErJc
+013308630-013311677	QmYHyegTZYecms9P9gCuNUp1vBZ7XHQohswt8iqmdmNhCe	QmP81iaNY88aHYaFrJPST1cT8gD76q3ZtnCHBk1YJA4dM6
+`
+
+func TestReadTabManifest(t *testing.T) {
+	m, err := ReadTabManifest(strings.NewReader(tabManifestSource))
+
```
tjayrush: Remove line

src/apps/chifra/pkg/pinlib/manifest/json_manifest_test.go
```diff
@@ -0,0 +1,120 @@
+package manifest
+
+import (
+	"strings"
+	"testing"
+)
+
+var manifestSource = `
+{
+  "fileName": "pin-manifest.json",
```
tjayrush, line 22: This data structure will probably change and made much less complicated, but I will do that only after we merge this PR.

src/apps/chifra/pkg/pinlib/manifest/json_manifest_test.go
```diff
@@ -0,0 +1,120 @@
...
+    }
+   ],
+  "prevBlockRange": "",
+  "prevPins": []
+}
+`
+
+func TestReadManifest(t *testing.T) {
+	m, err := ReadJSONManifest(strings.NewReader(manifestSource))
+
```
tjayrush: Remove line

src/apps/chifra/pkg/pinlib/chunk/errors.go
```diff
@@ -0,0 +1,42 @@
...
+func (err *ErrSavingCorruptedDownload) Error() string {
+	return fmt.Sprintf("file corrupted: %s (expected %d but got %d bytes)", err.fileName, err.expected, err.read)
+}
+
+type ErrSavingCannotDecompress struct {
+	fileName string
+	reason   error
+}
+
+func (err *ErrSavingCannotDecompress) Error() string {
```
tjayrush, line 34: Maybe we could come up with a way to handle all errors strings while we're still early. Like should all error strings be at the top of the file? Or in its own file?

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+	"time"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/pinlib/manifest"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/sigintTrap"
+	"github.com/panjf2000/ants/v2"
+)
+
+// The size of worker pools for downloading and saving data
+var poolSize = runtime.NumCPU() * 2
```
tjayrush: This value should be a configuration file item. Also, we should try to figure out a way to test the performance here. I've seen some code that would suggest NumCPU should be divided by 2, not doubled. Not sure...

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+
+type ChunkType uint
+
+const (
+	BloomChunk ChunkType = iota
+	IndexChunk
+)
+
+type ProgressEvent uint
+
```
tjayrush: Remove lines between type def of an enum and the list of enums.

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/pinlib/manifest"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/sigintTrap"
+	"github.com/panjf2000/ants/v2"
+)
+
+// The size of worker pools for downloading and saving data
+var poolSize = runtime.NumCPU() * 2
+
+type ChunkType uint
+
```
tjayrush, line 39: Remove lines between type def of an enum and the list of enums.

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+	return &fetchResult{
+		body:      body,
+		totalSize: fileSize,
+	}, nil
+}
+
+// For each pin in pins, downloads, unzips and saves the chunk of type indicated by chunkType.
+// Progress is reported to progressChannel.
+func GetChunksFromRemote(pins []manifest.PinDescriptor, chunkType ChunkType, progressChannel chan<- *ChunkProgress) {
```
tjayrush: Instead of having a special `ChunkProgress` type, can we have some sort of generic progress channel so we can use it elsewhere? It feels like we want to be consistent in progress reporting so the front end can report on any progress from the backend.

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+	var downloadWg sync.WaitGroup
+	var writeWg sync.WaitGroup
+
+	defer func() {
+		cancel()
+	}()
+
+	gatewayUrl := config.ReadBlockScrape().Dev.IpfsGateway
+
+	pool, err := ants.NewPoolWithFunc(poolSize, func(param interface{}) {
```
tjayrush: Can we rename this `pool` to `downloadPool` to make the following code clearer?

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+					Message: err.Error(),
+				}
+			}
+
+			if gatewayUrl == config.DefaultIpfsGateway {
+				time.Sleep(time.Second * 5) // Align with Pinata rate limit
+			}
+		}
+	})
+	defer pool.Release()
```
tjayrush: Rename to `downloadPool`.

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+		}
+		// Range over a channel is blocking, so the following line will be executed only
+		// when we finished saving pins
+		writeWg.Done()
+	}()
+
+	pinsToDownload := FilterDownloadedChunks(pins, cacheLayout)
+	for _, pin := range pinsToDownload {
+		downloadWg.Add(1)
+		pool.Invoke(pin)
```
tjayrush: rename to `downloadPool`.

src/apps/chifra/pkg/config/configuration.go
```diff
@@ -0,0 +1,42 @@
+package config
+
+var DefaultIpfsGateway = "http://gateway.ipfs.io/ipfs/"
```
tjayrush: This should be `https://ipfs.unchainedindex.io/ipfs`.

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+			} else {
+				progressChannel <- &ChunkProgress{
+					Pin:     &pin,
+					Event:   ProgressError,
+					Message: err.Error(),
+				}
+			}
+
+			if gatewayUrl == config.DefaultIpfsGateway {
+				time.Sleep(time.Second * 5) // Align with Pinata rate limit
```
tjayrush: Do we really need to sleep for 5 seconds each time we hit the endpoint? That seems high. I think if we use the dedicated endpoint we can remove this or lessen it significantly.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func runPins(cmd *cobra.Command, args []string) {
-	options := ""
-	if PinsOpts.list {
-		options += " --list"
+func handleList(format string) {
+	// Load manifest
+	manifestData, err := manifest.FromLocalFile()
```
tjayrush: If the --remote option is on, we can pull the manifest from the smart contract and (maybe optionally) store in locally. Or--broader question--how does the local manifest.txt file get updated?

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		logger.Fatal("Cannot open manifest file:", err)
```
tjayrush: "Cannot open local manifest file..."

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+	}
+
+	// Sort pins
+	sort.Slice(manifestData.NewPins, func(i, j int) bool {
+		iPin := manifestData.NewPins[i]
+		jPin := manifestData.NewPins[j]
```
tjayrush: No need to parse this. You can simply sort the string which is why the filenames have leading zeros.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+
+		jNumber, err := strconv.ParseInt(jKey, 10, 64)
+		if err != nil {
+			return false
+		}
+
+		return iNumber < jNumber
+	})
+
+	// dszlachta: I think we should move printing in the given format to a package
```
tjayrush: Yes. I agree.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+		}
+
+		fmt.Println(string(marshalled))
+
+		return
 	}
-	if PinsOpts.init {
-		options += " --init"
+
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
```
tjayrush: The tab writer is weird. If I do

```
NewWrite(os.Stdout, 0, 0, 1, '|', 0)
```

and the print three items ten characters long on one row and three items two characters long I get multiple '|' between fields, so this produces a 'table' not a tab delimited file which is un-expected.

Maybe I expected the wrong thing from a tabWriter. Maybe it's intended to write an aligned table using spaces...

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+	fmt.Fprintf(writer, "filename\tbloomhash\tindexhash\n")
+	for _, pin := range manifestData.NewPins {
+		fmt.Fprintf(writer, "%s\t%s\t%s\n", pin.FileName, pin.BloomHash, pin.IndexHash)
```
tjayrush: But then why use tabs here? Confused.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+	progressToLabel := map[chunk.ProgressEvent]string{
+		chunk.ProgressDownloading: "Unchaining",
+		chunk.ProgressUnzipping:   "Unzipping",
+		chunk.ProgressValidating:  "Validating",
+		chunk.ProgressError:       "Error",
 	}
+
+	var pinsDone uint
+
+	for event := range progress {
```
tjayrush: This is elegant code. I really like it.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+			logger.Log(logger.Info, pinsDone, "pin(s) were (re)initialized")
+			break
+		}
+
+		if event.Event == chunk.ProgressError {
+			failed = append(failed, *event.Pin)
+		}
+
+		if event.Event == chunk.ProgressDownloading {
+			logger.Log(logger.Info, "Unchaining", chunkTypeToDescription[chunkType], event.Message, "to", event.Pin.FileName)
```
tjayrush: Can we produce logging info that doesn't include a new line (use "\r" instead) so it 'spins'?

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+		}
+
+		if event.Event == chunk.ProgressDone {
+			pinsDone++
+			continue
+		}
+
+		eventLabel := progressToLabel[event.Event]
+
+		logger.Log(logger.Info, eventLabel, event.Pin.FileName, event.Message)
```
tjayrush: Can we produce logging info that doesn't include a new line (use "\r" instead) so it 'spins'?

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+func printManifestHeader() {
+	// The following two values should be read from manifest.txt, however right now only TSV format
+	// is available for download and it lacks this information
+	logger.Log(logger.Info, "hashToIndexFormatFile:", "Qmart6XP9XjL43p72PGR93QKytbK8jWWcMguhFgxATTya2")
+	logger.Log(logger.Info, "hashToBloomFormatFile:", "QmNhPk39DUFoEdhUmtGARqiFECUHeghyeryxZM9kyRxzHD")
+	logger.Log(logger.Info, "unchainedIndexAddr:", pinlib.GetUnchainedIndexAddress())
+	logger.Log(logger.Info, "manifestHashEncoding:", pinlib.GetManifestHashEncoding())
+}
+
+func handleInit() {
```
tjayrush: How hard would it be to separate `pinMan` into a separate folder? I like to put different functions of the command line tools into different files and call them handle_whatever. So there would be a folder for `pinMan` with `handle_list.go` and `handle_init.go` and so forth. Otherwise, I can see this `cmd` folder growing into hundreds of files.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+
+		if len(failedChunks) > 0 {
+			retry(failedChunks, 3, func(pins []manifest.PinDescriptor) []manifest.PinDescriptor {
+				logger.Log(logger.Info, "Retrying", len(pins), "bloom(s)")
+				return downloadAndReportProgress(pins, chunkType)
+			})
+		}
+	}
+
+	go func() {
```
tjayrush: This is so sweet. Really nice code here. I love it. I don't really understand it, but it's lovely

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+		handleList(RootOpts.fmt)
+		return
+	}
+	if PinsOpts.init || PinsOpts.freshen {
+		printManifestHeader()
+		handleInit()
+		return
+	}
+	if PinsOpts.remote {
+		logger.Fatal("Not implemented")
```
tjayrush, line 96: This option doesn't need to be 'implemented'. It's an option to apply to the `--list` option that instructs the list option where to get the manifest. By default, `--list` reports on the local manifest but this option says report on the `--remote` manifest (i.e. the one pointed to by the smart contract).

We may also need an option to report on all pins on Pinata.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
 	}
 	if PinsOpts.sleep != .25 {
-		options += " --sleep " + fmt.Sprintf("%.1f", PinsOpts.sleep)
+		logger.Fatal("Not implemented")
```
tjayrush, line 99: The `--sleep` option was intended to give the end-user control over the sleep needed to avoid timeouts on the servers.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,226 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
 	}
 	if PinsOpts.share {
-		options += " --share"
-	}
-	arguments := ""
-	for _, arg := range args {
-		arguments += " " + arg
+		logger.Fatal("Not implemented")
```
tjayrush, line 102: The `--share` option was intended to allow the user to participate in sharing the index via IPFS. It requires either an 'IPFS client' (similar to the EthClient) or a running `ipfs daemon`.

src/apps/chifra/cmd/pins_run_test.go
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * This program is free software: you may redistribute it and/or modify it under the terms
+ * of the GNU General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or (at your option) any later version. This program is
+ * distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * General Public License for more details. You should have received a copy of the GNU General
+ * Public License along with this program. If not, see http://www.gnu.org/licenses/.
+ *-------------------------------------------------------------------------------------------*/
+/*
```
tjayrush: Remove this comment. It's not true.

src/apps/chifra/cmd/pins_run_test.go
```diff
@@ -0,0 +1,64 @@
+/*-------------------------------------------------------------------------------------------
```
tjayrush: We need copyright text at the top of every file.

src/apps/chifra/pkg/pinlib/directories.go
```diff
@@ -0,0 +1,73 @@
...
+	subdirectories := []string{
+		"blooms",
+		"finalized",
+		"ripe",
+		"staging",
+		"unripe",
+	}
+
+	_, err := os.Stat(indexPath)
+
```
tjayrush: No blank line here.

src/apps/chifra/pkg/pinlib/directories.go
```diff
@@ -0,0 +1,73 @@
...
+	if os.IsNotExist(err) && indexPath != config.DefaultIndexPath {
+		return &ErrCustomizedPath{
+			indexPath: indexPath,
+		}
+	}
+
+	// directory doesn't exists, but it is the default one: create it
+	if os.IsNotExist(err) {
+		err = os.Mkdir(indexPath, 0755)
+
```
tjayrush, line 61: No blank line here.

src/apps/chifra/pkg/pinlib/directories.go
```diff
@@ -0,0 +1,73 @@
...
+			return err
+		}
+	}
+
+	// make sure all subdirectories exist as well, or create the ones that
+	// are missing
+	for _, subdirectory := range subdirectories {
+		subdirPath := path.Join(indexPath, subdirectory)
+		_, err = os.Stat(subdirPath)
+
```
tjayrush: No blank line here.

src/apps/chifra/cmd/pins_run.go
```diff
@@ -66,32 +81,238 @@ func validatePinsArgs(cmd *cobra.Command, args []string) error {
...
+
+func runPins(cmd *cobra.Command, args []string) {
+	if PinsOpts.list {
+		printManifestHeader()
+		handleList(RootOpts.fmt)
+		return
+	}
+	if PinsOpts.init || PinsOpts.freshen {
+		printManifestHeader()
+		err := pinlib.EstablishDirectories()
```
tjayrush, line 80: This works, but I think it should happen at line 287 above. In that way, if -- for some reason -- the user does `chifra pins --list` before `chifra pins --init` they won't get an inelegant error message. In other words, it's a prerequisite that the folders exist as they should prior to any pinMan command. Perhaps this function should be in the `init` function, but that's a problem because that code is auto-generated. That the purpose of `// EXISTING_CODE` blocks in the auto-generated code. Let's not worry about it for now. We can fix this once the PR is merged.

src/apps/chifra/go.mod
dszlachta: Right. I don't think why it happens, but `go mod tidy` changes it in the right way... We can have a commit hook to call `go mod tidy` automatically

src/apps/chifra/pkg/rpcClient/client.go
dszlachta: My idea was that because we have to import this module (`rpcClient`) whenever we want to do RPC, it would be easier to have this helper functions (that I assume we will be using often) in the same "namespace", so we don't have to import two packages (our `rpcClient` and go-ethereum's `common`)

src/apps/chifra/pkg/rpcClient/client.go
dszlachta: 👍

src/apps/chifra/pkg/blockRange/range_test.go
dszlachta: Sorry, I missed a few spots and left "my" formatting. Should be fixed now.

src/apps/chifra/pkg/config/utils.go
dszlachta: Sure

src/apps/chifra/pkg/config/utils.go
dszlachta: Done

src/apps/chifra/pkg/config/types.go
dszlachta: Good catch, thanks. But should it be `Balance` or `Balances`? C++ uses the latter

src/apps/chifra/pkg/config/configuration.go
dszlachta: Sure thing 👍

src/apps/chifra/pkg/config/configuration.go
dszlachta: On line 33 we create a new struct and fill it with the default values. When we read values from a configuration file and a value for a given key is missing, go will just use the default that we set up earlier

src/apps/chifra/pkg/file/lock.go
dszlachta: According to fcntl (2) manual:
```
short l_whence;  /* How to interpret l_start:
                        SEEK_SET, SEEK_CUR, SEEK_END */
```
So I gather that `whence` is not related to the size of a file, but it's rather just a flag. It's int16 because that's what golang wants...

src/apps/chifra/pkg/logger/logger.go
dszlachta: Done

src/apps/chifra/pkg/pinlib/manifest/range.go
dszlachta: An excellent point.
I saw `x_2_y` somewhere in go-ethereum, but have never seen it in golang docs (which is not a problem I guess). Let me know if you'd like to go with `x_2_y`, in the meantime I will change this name to `StringToManifestRange`

src/apps/chifra/pkg/pinlib/manifest/range.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	parts := strings.Split(source, "-")
+
+	start, err := strconv.ParseUint(parts[0], 10, 64)
+	if err != nil {
+		return mr, err
+	}
+
+	end, err := strconv.ParseUint(parts[1], 10, 64)
```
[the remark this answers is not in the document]
dszlachta: Would it be a valid manifest range then?
The function will report an error, because my understanding is that, as opposed to block ranges, manifest range is always `from-to`

src/apps/chifra/pkg/pinlib/save.go
dszlachta: But we need to `Write()` or `WriteAll()` first to have anything returned from `Error()`

src/apps/chifra/pkg/pinlib/contract.go
dszlachta: Sure, should I add it? Which file would it be?

src/apps/chifra/pkg/pinlib/contract.go
dszlachta: Good point. 
If the contract doesn't contain `manifestHash` then I guess the only thing we can do is to `Fatal` or `panic` with a nice message for the user

src/apps/chifra/pkg/pinlib/manifest/manifest.go
dszlachta: Not sure either. Maybe they need `[]byte` more often, because of sending data to RPC and decoding/encoding hexes and addresses, so they don't want to type cast?

src/apps/chifra/pkg/pinlib/manifest/manifest.go
dszlachta: Sorry, I wasn't aware of it

src/apps/chifra/pkg/pinlib/manifest/tab_manifest.go
dszlachta: Hm, this would be an empty file, right? Maybe we should have a validator for manifest file...
I changed this function to check `len` and return error if it's 0

src/apps/chifra/pkg/pinlib/chunk/errors.go
dszlachta: It's a nice idea, but because all of these strings need some data (like file names, etc) I guess we would have to put them into functions... and this would make the code a bit more complex I think.

src/apps/chifra/pkg/pinlib/chunk/chunk.go
dszlachta: I can put it into a configuration file, but which one?

We could use hyperfine to measure performance of, say, `NumCPU() * 2`, `NumCPU() * 4` and `NumCPU() / 2`

tjayrush: For better or worse, each command has it's own configuration file. And that file gets merged into the global config file. So all the values in the global file get read first, and then all the values in whatever command's config file get laid on top of that. So this value can be either in the global config (trueblocks.toml) or the command's local config (in this case, pinMan.toml).

Although, on second thought, because there are a few other config items below (the smart contract address and the function call) and because I need this config item and both of those in `blockScrape` and because you already have a `blockScrape` config, I think we should use that.

An alternative would be config items in a [section] of the trueblocks.toml file, but then we have a migration issue.

src/apps/chifra/pkg/pinlib/chunk/errors.go
tjayrush: This is an example of some code I made a note of that I thought did the 'error thing' well: https://github.com/storj/uplink/blob/v1.7.0/bucket.go#L19. Maybe we can copy this idea.

src/apps/chifra/pkg/pinlib/manifest/manifest.go
tjayrush: We'll have to think about this, I guess.

src/apps/chifra/pkg/pinlib/manifest/manifest.go
tjayrush: That's fine. Once we merge we can fix it. The existing code uses the simple tab separated file, so we can worry about this once it's merged.

src/apps/chifra/pkg/pinlib/save.go
tjayrush: Oh. I see. Never mind. Closing...

src/apps/chifra/pkg/rpcClient/client.go
tjayrush: Make sense. I like have out own control of it too, so we'll leave it.

src/apps/chifra/pkg/pinlib/chunk/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+// For each pin in pins, downloads, unzips and saves the chunk of type indicated by chunkType.
+// Progress is reported to progressChannel.
+func GetChunksFromRemote(pins []manifest.PinDescriptor, chunkType ChunkType, progressChannel chan<- *ChunkProgress) {
+	// Downloaded content will wait for saving in this channel
+	writeChannel := make(chan *jobResult, poolSize)
+	// Context lets us handle Ctrl-C easily
+	ctx, cancel := context.WithCancel(context.Background())
+	cacheLayout := &CacheLayout{}
+	cacheLayout.New(chunkType)
+	var downloadWg sync.WaitGroup
```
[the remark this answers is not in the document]
dszlachta: My thinking was that download takes a different amount of time than writing to the disk. So I didn't want to block writing jobs by waiting for download to complete. This code downloads chunks and push them to a channel, not waiting for write. Write workers just take the data from the channel and write it in parallel whenever a worker is free.

Because downloading, unzipping and validating happens in-memory, only valid files will be written to the destination directory.

src/apps/chifra/go.mod
tjayrush: https://github.com/TrueBlocks/trueblocks-core/issues/1855

src/apps/chifra/cmd/pins_run.go
tjayrush: https://github.com/TrueBlocks/trueblocks-core/issues/1856

src/apps/chifra/cmd/pins_run.go
tjayrush: https://github.com/TrueBlocks/trueblocks-core/issues/1857

src/apps/chifra/cmd/pins_run.go
tjayrush: This is documented in issue #1856

src/apps/chifra/cmd/pins_run.go
tjayrush: We can wait to do this until later. Made an issue: https://github.com/TrueBlocks/trueblocks-core/issues/1859

src/apps/chifra/cmd/pins_run_test.go
tjayrush: https://github.com/TrueBlocks/trueblocks-core/issues/1860

src/apps/chifra/pkg/config/configuration.go
tjayrush: Sounds good.

src/apps/chifra/pkg/file/lock.go
tjayrush: Thanks. No action needed here.

PR #1865, opened 2021-11-02
WIP: Bugfix/pins test fixes (not ready for merge)

tjayrush:
I make this PR just so you can start reviewing it. It doesn't change much but it points out a few failing test cases (which I turned off for now so I can post this).

To run tests failed test, edit `pinMan.csv` in the `./src/dev_tools/testRunner/testCases/apps` folder to change all any test with 'cmd' under `cmd` (but only if `post` == 'y') to `both`.

The run `~/.local/bin/chifra/test/test-all.sh apps/pinMan/` from the `./build` folder.

You should see a number of failed tests which I documented in the `pinMan.csv` file comments.

You don't have to fix them. I have a good idea why they're failing, but if you want to look at the tests that fail because of not sending the 'error' string on stdout, you can look at that one.

I'll pick this back up in the morning.


src/apps/chifra/pkg/utils/utils.go
```diff
@@ -57,9 +57,9 @@ func TestLogArgs(name string, args []string) {
 		return
 	}
 
-	fmt.Fprintf(os.Stderr, "TIME ~ CLOCK - <INFO>  : %s\n", name)
+	fmt.Fprintf(os.Stderr, "INFO[DATE|TIME] %s\n", name)
```
dszlachta: I think we should use logger.Log to be sure that we have consistent message format across the whole codebase

src/apps/chifra/cmd/pinman/handle_list.go
```diff
@@ -32,29 +34,43 @@ func HandleList(format string) {
...
+
+	if format == "" || format == "none" {
+		if utils.IsApiMode() {
+			format = "api"
+		} else {
+			format = "txt"
+		}
+	}
+
+	outFmt := "%s\t%s\t%s\n"
```
dszlachta, line 56: Was there any problem with `Table`? 
Printing strings with tabs between them doesn't guarantee a tabular output and this is what happens in C++ (I have header and some rows misaligned with the rest). I thought that it was a bug and `Table` was a way to fix it. But right now I think that maybe we want TSV output and not a table? In such a case, let's remove Table and have CSV writer in `output` package. We can then configure this CSV writer to use tabs as delimeters (just like in `manifest` package)

dszlachta (approved):

dszlachta:
Regarding writing errors to stderr when in API_MODE - I think we could add this to the logger (as logger.Error). What do you think?

src/apps/chifra/cmd/pinman/handle_list.go
tjayrush: Yes. We want a file that can be imported into MS Excel.

But....now that I think about it....in the C++ code it can tell if it's being written to the console or not. If it is, it displays colored text, but if it's being re-directed (through `>`) or piped (through '|') we don't produce colored text. We could do the same thing here. I like the table view (with aligned columns), but it's not useful if redirected.

This is the line that knows if it's running against a terminal: https://github.com/TrueBlocks/trueblocks-core/blob/master/src/libs/utillib/options_base.cpp#L64.

Here's an article about it in golang: https://www.darkcoding.net/software/in-go-is-your-terminal-a-console-or-a-pipe-isatty-golang/

src/apps/chifra/pkg/utils/utils.go
tjayrush: Nice catch.

src/apps/chifra/cmd/pinman/handle_list.go
dszlachta: Yeah, we could do that. But I think that we should at least keep format-specific logic in `output` package and use functions from there (to maintain consistency easily). For example, we could have `output.PrintTab`

src/apps/chifra/pkg/utils/utils.go
tjayrush: I agree.

src/apps/chifra/cmd/pinman/handle_list.go
tjayrush: I think we may want to use an interface here in a way similar to the way `CBaseClass` works in the C++ code. So we can send into any `CArchive` and `CBaseClass` and it will handle it correctly. `CArchive` behaves exactly like a file in the C++ code. So, for example, I can do (with any data type):

```
cout << item << endl; // write to stdout
cout << item << endl; // write to stderr

ostringstream os;
os << item << endl; // to write to string

CArchive archive;
archive.Open(...)
archive << item << endl; // to write to file

archive >> different_item; // to read
```

tjayrush:
> Regarding writing errors to stderr when in API_MODE - I think we could add this to the logger (as logger.Error). What do you think?

The trouble is that `cobra` prints the error message before we get a chance to intercept it for `invalid flags`. So, if we do `chifra -pins --list -fmt junk` (which has all valid flags, but invalid option to the `--fmt` flag), we can intercept that, but if we do `chifra pins --junk` we don't get a chance to decide between `stdout` and `stderr`.

dszlachta:
Ah, I see. Maybe we could try `command.SetErr` to direct the output to the right place: https://github.com/spf13/cobra/blob/eb3b6397b1b5d1b0a2cd66a9afe0520f480c0a87/command.go#L254

Dawid Szlachta
2021年11月2日 11:19 +0100、Thomas Jay Rush ***@***.***> のメール:

tjayrush:
Can you play around with that? There are four or five test cases with `mode` set to `cmd` and `post` set to `y` in `~/src/dev_tools/testRunner/testCases/apps/pinMan.csv`. You can change the `mode` to `both` and the errors will appear.  (Don't change the value of `mode` if `post` == `n`.)

You can run only the `pinMan` tests with `~/.local/bin/chifra/test/test-all.sh apps/pinMan`

tjayrush:
Merging this since it's at least as good as current code.

PR #1866, opened 2021-11-02
Replaced custom config package with viper

dszlachta:
- Added Viper
- Set defaults for configuration values
- Enabled reading environment variables (e.g. `[dev] ipfs_gateway` -> `DEV_IPFS_GATEWAY=mygateway.example.com`)
- Loading config files into structs to make them first-class citizens in our code


src/apps/chifra/pkg/config/block_scrape.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// ReadBlockScrape reads the configuration located in blockScrape.toml file
+func ReadBlockScrape() *BlockScrape {
+	if !blockScrapeRead {
+		MustReadConfig(blockScrapeViper, &cachedBlockScrape)
+
+		// Validate the URL to ensure we have it in the correct format, so that ethClient.Dial
+		// will not panic
+		if !strings.HasPrefix(cachedBlockScrape.Dev.IpfsGateway, "http") {
```
tjayrush, line 69: I usually also check for an ending `/' whenever I read from a config file. If it's not there, I add it. In this way I always know that any config file path or URL only needs the subfolder or route to be added and I don't have to check for the trailing 'slash'.

src/apps/chifra/pkg/config/block_scrape.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type BlockScrape struct {
+	Requires       blockScrapeRequires
+	Dev            blockScrapeDev
+	UnchainedIndex blockScrapeUnchainedIndex
+}
+
+// init sets up default values for the given configuration
+func init() {
```
tjayrush, line 51: So, just for my information. Is it true that a `init` function in any file will get called prior to startup? In other words, prior to `main`, so we can count on these values being here before main?

src/apps/chifra/pkg/config/global_config.go
```diff
@@ -12,7 +12,14 @@
  *-------------------------------------------------------------------------------------------*/
```
tjayrush: Can we please rename this file `trueblocks.go`? In that way, when I look in this folder, each file will be named the same as it's subcommand route except for this one. And the names of these go files will reflect the actual names of the toml files they read.

src/apps/chifra/pkg/config/utils.go
```diff
@@ -64,29 +65,20 @@ func GetConfigPath(fileName string) string {
...
-		fileNameWithExt += ".toml"
-	}
-
-	filePath := GetConfigPath(fileNameWithExt)
+// MustReadConfig calls v's ReadInConfig and fills values in the
+// given targetStruct. Any error will result in a call to logger.Fatal
+func MustReadConfig(v *viper.Viper, targetStruct interface{}) {
+	v.AddConfigPath(GetConfigPath(""))
+	v.AutomaticEnv()
+	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
```
tjayrush, line 74: See the viper usages in `./src/go-apps/blaze/cmd/options.go` (I think). We prepend `TB_` in front of the environment variables. In this way, they are all easily discoverable with `env | grep TB_` from the command line. But don't change the settings in the files, just the environment variables.

tjayrush:
This looks great. Just close these few issues and we can merge it.

src/apps/chifra/pkg/config/block_scrape.go
dszlachta: This is exactly why the code later uses `path.Join`, which will add ending `/` if it's missing

src/apps/chifra/pkg/config/block_scrape.go
dszlachta: Yes, it's described here (5th paragraph from the end of the section): https://golang.org/ref/spec#Package_initialization

src/apps/chifra/pkg/config/global_config.go
dszlachta: Definitely, good idea

tjayrush (commented):
I forgot to submit this one comment...

PR #1870, opened 2021-11-02
Bugfix/go api errors

tjayrush:
This solves the issue we discussed about returning 'errors' to the stdout when in API mode, but it's not that well done. I'm open to suggestions. It does make all but one test pass.

That test is a the special case we discussed when `cobra` handles invalid, unknown flags and doesn't give our code a chance to respond. That test is turned off with `cmd` in the mode column for `pinMan.csv`.

We discussed fixing that by moving the server code into `chifra`.

If you'd like, please review this, otherwise, please go ahead and merge it.

### DIFF
--- a/src/apps/chifra/cmd/pins/handle_list.go
+++ b/src/apps/chifra/cmd/pins/handle_list.go
@@ -53,8 +53,6 @@ func HandleList() {
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


dszlachta (approved):
Looks good 👍

Issue #1873, opened 2021-11-04
chifra abis --find with many searches fails


tjayrush:
Search for `// TODO: see issue #1873`

Uncomment the `wg.Done()` and `return` lines. Comment out the `os.Exit()` line

Run this command:

```
[...]

PR #1874, opened 2021-11-04
chifra abis --find

tjayrush:
Simple movement of code out of `./src/go-apps` folder. Also, allow multiple .sol files on command line for `--sol` option.


src/apps/chifra/cmd/abis/handle_find.go
```diff
@@ -13,12 +13,106 @@
...
+			if err != nil {
+				fmt.Printf("Couldn't parse argument %s\n", arg)
+			} else {
+				arguments = append(arguments, asBytes)
+			}
+		}
+	}
+
+	var wg sync.WaitGroup
+	checkOne, _ := ants.NewPoolWithFunc(runtime.NumCPU()*2/3, func(input interface{}) {
```
dszlachta: Pool size is a configuration value now. You can get it by calling `config.ReadBlockScrape().Dev.MaxPoolSize`

src/apps/chifra/cmd/abis/handle_find.go
```diff
@@ -13,12 +13,106 @@
...
+		byts := []byte(str)
+		encBytes := crypto.Keccak256(byts)[:4]
+		for _, arg := range arguments {
+			if !testMode {
+				fmt.Fprintf(os.Stderr, "Scanning: %s\r", input)
+			}
+			if bytes.Equal(encBytes, arg) {
+				prettyPrint(hex.EncodeToString(encBytes), input)
+				// TODO: see issue #1873
+				// wg.Done()
```
dszlachta: Regarding issue #1873, this `Done()` is doubled,  because `defer wg.Done()` from line 55 will be always called before this function exits, so maybe this is why you were getting `negative WaitGroup counter`

dszlachta (approved):

src/apps/chifra/cmd/abis/handle_find.go
tjayrush: Done

src/apps/chifra/cmd/abis/handle_find.go
tjayrush: That fixed it. Thanks!

tjayrush:
@dszlachta When you get a chance, please go ahead and merge this. You're welcome to make a few comments if you have any and I will fix them, but this is at least as good as the previous code, even though it doesn't use channels right and it needs improvement. I need to study more about channels/go routines but in the meantime, the code passes all tests (in fact it found a bug in the old code).

PR #1880, opened 2021-11-06
Feature/move flame to apps

dszlachta:
- Moved go-apps/flame to chifra/serve
- Added logic to easily return json from other Go functions
- Changed `/pins` endpoint to call `pins` directly
- Removed unused code
- Fixed TrueBlocksConfig being still named "Global" in some places
- Added output.JsonFormatted to make sure we produce JSON output in one place only (so it's consistent across the whole app)
- Applied staticcheck fixed (replaced 1 deprecated feature, simplified a few places)


src/apps/chifra/cmd/serve/exec/admin_pins.go
```diff
@@ -0,0 +1,62 @@
...
+var ErrMultipleModes = errors.New("only one of ?list or ?init is supported")
+var ErrNoMode = errors.New("you must choose one of ?list or ?init")
+
+// AdminPins runs pin-operating functions depending on query parameters
+func AdminPins(request *http.Request) ([]manifest.PinDescriptor, error) {
+	query := request.URL.Query()
+	mode := ""
+	all := false
+
+	for key := range query {
```
tjayrush, line 21: I'm not sure this is what we want to do. As you know all of the 'command line' options code is programmatically generated from a .csv file. By putting this code here (as opposed to say centralizing it inside of the `pins` package, we've created a second (actually third) place where the auto-generation code will eventually have to look.

I know it seems a bit troublesome, but I cannot tell you have wonderful it is to have the auto-generation stuff. In the same way you can be confident that changes to your code don't get broken if you have good testing and that allows you to be ultra-flexible, being able to autogenerate code for all command-line options -- especially documentation -- allows for complete flexibility in changing command lines.

Upshot: can we somehow move this into either a common function that validates the command lines (even if this means refactoring the existing `validatePins` routine to take a string arrays, or, at the very least put this code into the package so it's all in one place (as opposed to in the `serve` subfolder?

src/apps/chifra/cmd/serve/exec/admin_pins.go
```diff
@@ -0,0 +1,62 @@
...
+	// Make sure all required directories exist
+	err := pinlib.EstablishDirectories()
+	if err != nil {
+		return nil, err
+	}
+	var responseBody []manifest.PinDescriptor
+	var responseError error
+
+	switch mode {
+	case "list":
```
tjayrush, line 46: If I add new option to the `pinMan` command (which I will be doing when I incorporate `chunkMan` functions) I will have to edit not only this code but also the `runPins` code in the `pins` package. Can we combine the two?

src/apps/chifra/cmd/serve/options.go
```diff
@@ -13,12 +13,15 @@ package utils
...
-var Green string = "\033[32m"
-var Yellow string = "\033[33m"
-var Blue string = "\033[34m"
-var Purple string = "\033[35m"
-var Cyan string = "\033[36m"
-var White string = "\033[37m"
-var Black string = "\033[30m"
+// OptionsType makes command line options available to the package
+type OptionsType struct {
+	Scrape  bool
```
tjayrush: I think (but I'm not sure) that both `Scrape` and `Monitor` can be removed from this structure. These are hold-overs from the old scraper code which was removed from the old server code a while ago.

src/apps/chifra/cmd/serve/progress_test.go
```diff
@@ -10,7 +10,7 @@
  * General Public License for more details. You should have received a copy of the GNU General
  * Public License along with this program. If not, see http://www.gnu.org/licenses/.
  *-------------------------------------------------------------------------------------------*/
-package server
+package serve
```
tjayrush, line 13: Nice catch.

src/apps/chifra/cmd/serve/routes.go
```diff
@@ -186,9 +187,15 @@ func AdminInit(w http.ResponseWriter, r *http.Request) {
 	CallOneExtra(w, r, "chifra", "init", "init")
 }
 
-// AdminPins help text todo
+// AdminPins handles /pins route
 func AdminPins(w http.ResponseWriter, r *http.Request) {
-	CallOneExtra(w, r, "chifra", "pins", "pins")
+	result, err := exec.AdminPins(r)
```
tjayrush, line 192: I LOVE THIS! I CAN'T WAIT TO SEE MORE OF IT!

src/apps/chifra/cmd/serve/server.go
```diff
@@ -0,0 +1,16 @@
+package serve
+
+import "net/http"
+
+// TODO: this should be soon part of pkg/utils
```
tjayrush, line 5: Yes. Or the package `colors`, or (maybe better) `colors` should be combined with `utils`. Your call.

src/apps/chifra/cmd/serve_run.go
```diff
@@ -13,12 +13,20 @@
...
 
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/cmd/serve"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/output"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/validate"
 	"github.com/spf13/cobra"
 )
 
+// TODO: this should be soon part of pkg/utils
+var Off string = "\033[0m"
```
tjayrush, line 27: Yes

src/apps/chifra/cmd/serve_run.go
```diff
@@ -27,14 +35,25 @@ func validateServeArgs(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func runServe(cmd *cobra.Command, args []string) {
-	options := ":8080"
-	if len(ServeOpts.port) > 0 {
-		options = " --port " + ServeOpts.port
+func PrintServeSettings() {
+	if ServeOpts.port != ":8080" {
```
tjayrush, line 39: I presume this is where you fix the bug. Did we have a test for this?

I don't think we did -- in fact, I doubt we had very many tests at all for the server code.

src/apps/chifra/pkg/output/json.go
```diff
@@ -20,29 +20,44 @@ import (
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/validate"
 )
 
-// PrintJson marshals its arguments and prints JSON in a standardized format
-func PrintJson(serializable interface{}) error {
-	var response map[string]interface{}
+type JsonFormatted struct {
```
tjayrush, line 23: I'm still a little bit behind you on how the Outputting stuff should work. We can talk about it on Monday.

src/apps/chifra/pkg/output/json.go
```diff
@@ -20,29 +20,44 @@ import (
...
-		}
+// AsJsonBytes marshals JsonFormatted struct, populating Meta field if
+// needed
+func AsJsonBytes(j *JsonFormatted) ([]byte, error) {
+	var result JsonFormatted
 
+	if Format == "json" {
+		result = *j
 	} else {
+		// TODO: global validate.Errors is not server-safe
```
tjayrush, line 37: Can you explain this to me? I'm sure it's true -- globals are almost always bad -- but how do you know?

src/apps/chifra/cmd/serve/exec/admin_pins.go
```diff
@@ -0,0 +1,62 @@
+package serveRoutes
```
tjayrush: I didn't know you could have a package named differently than the folder it's in. I'm not saying it's wrong, I'm just curious to see it.

src/apps/chifra/cmd/serve/exec/admin_pins.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"errors"
+	"net/http"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/cmd/pins"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/pinlib"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/pinlib/manifest"
+)
+
+var ErrMultipleModes = errors.New("only one of ?list or ?init is supported")
```
tjayrush, line 12: Aren't these duplicated in `pins_run.go`? It's okay if they are for now, but eventually, I'll want to consolidate these and bring them under the auto-code-generation tool. Perhaps you're already anticipating that?

src/apps/chifra/cmd/serve/exec/admin_pins.go
```diff
@@ -0,0 +1,62 @@
...
+var ErrNoMode = errors.New("you must choose one of ?list or ?init")
+
+// AdminPins runs pin-operating functions depending on query parameters
+func AdminPins(request *http.Request) ([]manifest.PinDescriptor, error) {
+	query := request.URL.Query()
```
tjayrush, line 17: Couldn't we

- convert this into a `cobra` command and/or
- convert the `cobra` command into an `http.Request` and/or
- convert both of them into a key/value map

so the validation and execution of each command could be the same code? Perhaps you already anticipate that?

src/go-apps/flame/utils/utils.go
```diff
@@ -1,75 +0,0 @@
-package utils
```
tjayrush, line 1: I'm very happy to see this duplicate code disappear.

tjayrush:
This is great. I reviewed it quite carefully, so I fully understand what it does. 

The only thing I saw that seemed wrong really is that it runs on port `:80` as opposed to `:8080`. My guess is that's an oversight.

See my note in the discord about our call tomorrow.

src/apps/chifra/cmd/serve/options.go
```diff
@@ -13,12 +13,15 @@ package utils
...
+	Monitor bool
+	Port    string
+	Verbose int
```
tjayrush, line 19: I think only the `--port` option needs to be here. Verbose is part of the rootCmd and the other options don't apply to the server.

src/apps/chifra/cmd/serve/exec/admin_pins.go
dszlachta: Good point. I will try to reuse the existing code we have for validating command line arguments

src/apps/chifra/cmd/serve/exec/admin_pins.go
dszlachta: Sure

src/apps/chifra/cmd/serve_run.go
dszlachta: It was fixed on line 55, actually. 
I think we don't have a test for it. It seems we only have 2 tests for serve

src/apps/chifra/pkg/output/json.go
dszlachta: Sure. I just wanted to have this code in one place, so we always get consistent format

src/apps/chifra/pkg/output/json.go
dszlachta: Let's say there's an error in this array. It was caused by client A connecting and doing something wrong. Then client B connects, new goroutine starts and takes care of B's request. But there is still error in the array, so B will get it even tough they didn't do anything wrong. 
I know that TrueBlocks are a local application, but these two clients could be user opening two tabs in their browser (or maybe our future Electron UI). Or someone using API.

src/apps/chifra/cmd/serve/exec/admin_pins.go
dszlachta: Yes, as far as I know package name doesn't have to be the same as folder's, but you can only have one package per folder

dszlachta: Well, I actually renamed the folder and forgot to rename the package. I will fix it now, thanks for catching it

src/apps/chifra/cmd/serve/exec/admin_pins.go
dszlachta: Yes, they're almost the same. I could have used the ones declared in `pins_run`, but they mention `--option`, which can be misleading when using API (it would be `?option`). So either we remove `--` from the errors or leave it and hope we won't confuse API consumers that much.

src/apps/chifra/cmd/serve/exec/admin_pins.go
dszlachta: I totally get your point and I agree. But there are some problems to solve first:
1. Cobra's `Command.Run` returns nothing, so how do we get the results?
2. Running as a server requires a slight change in the behaviour of the programme. For example, we shouldn't use `log.Fatal` in the command, because this would kill the whole server. 
3. Running as server switches the code from printing out results and errors to returning them (this is why I separated HandleList into List and HandleList. The latter prints output, the former does the job and returns data as variables)

dszlachta:
Fixed the default port issue

src/apps/chifra/cmd/serve/exec/admin_pins.go
tjayrush: I would argue to leave it for now. Once things settle down, error messages become something generated from the auto-generator. At that time, we can decide and easily change all `--` to `?`. Until then, just leave them all with `--`.

src/apps/chifra/cmd/serve/exec/admin_pins.go
tjayrush: There is a version of `cobra's` `Run` called `RunE` which returns an error. I've been thinking we may want to use that instead of `Run`. I think people use it to do 'extra' validation before running but perhaps after doing some processing. So, validate is for things that can be validated early. `RunE` is for stuff that can't be validated until later.

src/apps/chifra/cmd/serve/options.go
tjayrush: I think you can remove `Sleep` and `Verbose` here.

src/apps/chifra/cmd/serve_run.go
tjayrush: Serve is weird as it has no real purpose from the command line and can't be queries by itself in the server case, so what would there be to test. Plus -- the tests themselves use the server constantly, so I don't think we really need any.

src/apps/chifra/pkg/output/json.go
tjayrush: Got it completely. Thanks.

PR #1890, opened 2021-11-09
Port/move abis find fix tests

dszlachta:
This is far from being perfect, but it does fix all pin API tests (only `to_file` are failing, but this is because we don't have this functionality yet).

I'm more than happy to continue working on it, but don't want to block you, hence the PR. Please feel free to merge it if it's OK or I will just try to make it better tomorrow.


tjayrush:
I'm making a few notes which I will attach to issues as soon as this is merged into the `develop` branch.

Issue #1881, opened 2021-11-07
Test fail if there's no blockScrape.toml file (which is the case on first install)


tjayrush:
If I go to a brand new machine and install and build TrueBlocks and run the tests, all the `pinMan` and all the `grabABI` tests fail.

This is easy to see why. The file `blockScrape.toml` does not exist and the MustReadConfig function insists that it does.

While MustReadConfig is correct for the main `trueBlocks.toml` file, it is not true for any other config file. In general, if a config file does not exist, the reading of the config file should not fail and the default values for the config should be returned in every case.

And, now on second thought, this should also be true of the `trueBlocks.toml` file.

Instead of insisting on its existence, we should allow it to default to regular defaults and allow it to report failure because the RPC endpoint doesn't exist for example.

tjayrush:
@dszlachta Can you make this issue a priority? On brand new installs, this is broken because the `blockScrape.toml` file is required but does not exist. I'd like to change it (and all config files for that matter) to simply use the default values if the file does not exist. (In other words, no config files (include `trueBlocks.toml`) MUST exist.)

dszlachta:
Sure

tjayrush:
Also -- every read of a path from any configuration file must ensure that there is a trailing slash. The user should be able to put both the trailing slash and not the trailing slash and we should not fail. As it works now, if the user does not add a trailing slash all hell breaks loose.

dszlachta:
Wow, that’s super strange, it shouldn’t fail like that. I will fix it.

tjayrush:
That may be on the develop branch which may not have your fix for that. I'll close this if it works once i merge your PR.

tjayrush:
I noticed on my Ubuntu box that there's a file called `blockSrape.toml` (misspelling). I'm not finding that string anywhere in the code, so it may have been fixed.

I am testing the PR now, so will let you know and close this issue if the PR fixes the problem.

PR #1894, opened 2021-11-10
Bugfix/1881 allow missing configs

dszlachta:
As described in #1881:
- Added option to make certain config files optional
- Marked trueBlocks.toml as optional as well and added all required defaults


tjayrush (commented):
Look good

PR #1900, opened 2021-11-16
Removed unused --transactions flag from acctExport

dszlachta:
I removed the flag from `cmd-line-options.csv` and then regenerated sources. I added files that have been changed. 
The option seemed unused (just as the compiler reported)


tjayrush:
Looks good

PR #1925, opened 2021-11-23
Removes underbar paramenter and type names from OpenAPI.yaml

tjayrush:


docs/content/api/openapi.yaml
```diff
@@ -801,7 +801,7 @@ paths:
             type: array
             items:
               type: string
-              format: tx_id_t
+              format: tx_id
```
dszlachta, line 804: Should this be camelCased as well? `nonZero` is.

dszlachta (approved):

docs/content/api/openapi.yaml
tjayrush: Yes. Can you do me a favor though? Since we've shut off generating the openapi.yaml file, just edit it directly (it appears in various places). I would have had to turn off a lot of test cases for this one. Once we bring your changes to the OpenAPI back into the auto-gen code (makeClass), I will fix it then.

dszlachta: Sure thing

PR #1930, opened 2021-11-25
Fixing meta

tjayrush:


src/apps/chifra/pkg/blockRange/range.go
```diff
@@ -73,6 +74,14 @@ type BlockRange struct {
 	Modifier     Modifier
 }
 
+func (br BlockRange) String() string {
```
dszlachta: This is actually producing a JSON string, right? Maybe we want to call this function `MarshallJSON` and keep `String` for a potential future use (e.g. at some point we may want to serialize BlockRange to `0-10000`)

dszlachta (approved):

src/apps/chifra/pkg/blockRange/range.go
tjayrush: Changing now.

tjayrush: Should it take a pointer to the object or the object itself?

dszlachta:
It can be a pointer, since 1) you don’t modify the object, 2) ranges are not large
See https://go.dev/doc/faq#methods_on_values_or_pointers

Dawid Szlachta

tjayrush:
> 

This helps: "Whether to define the receiver as a value or as a pointer is the same question, then, as whether a function argument should be a value or a pointer."

But in C++, I would use a constant reference as a parameter in this case. I don't think there is such as thing as a constant in Go. Is there?

tjayrush:
This helps too: "Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used"

dszlachta:
No, there’s no such constants in Go

Dawid Szlachta

tjayrush:
You can merge this if you think it's okay.

PR #1942, opened 2021-12-03
Adding a few things to the chifra pkgs folder

tjayrush:
This is just a few small things I added for the GitCoin stuff. Please review, and if you think it's okay, merge it. Thanks.


src/apps/chifra/pkg/tslib/tslib.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"encoding/binary"
+	"errors"
+	"os"
+	"sort"
+	"strings"
+	"time"
+)
+
+const tsPath = "/Users/jrush/Development/trueblocks-index/wild.0/unchained/ts.bin"
```
dszlachta: This path shouldn't be hardcoded, I guess

src/apps/chifra/pkg/tslib/tslib.go
```diff
@@ -0,0 +1,143 @@
...
+	})
+
+	if uint64(memory[index].Ts) != ts {
+		return &Timestamp{}, errors.New("timestamp not found")
+	}
+
+	return &memory[index], nil
+}
+
+func fromBn(bn uint64) (*Timestamp, error) {
```
dszlachta, line 98: What's `Bn`? Maybe we could change this name to something with a clearer meaning?

src/apps/chifra/pkg/tslib/tslib.go
```diff
@@ -0,0 +1,143 @@
...
+	return uint64(ret.Bn), err
+}
+
+func TsFromBn(bn uint64) (uint64, error) {
+	ret, err := fromBn(bn)
+	return uint64(ret.Ts), err
+}
+
+func TsFromDate(date string) (uint64, error) {
+	t, err := time.Parse("2006-01-02 15:04:05", strings.Replace(date, "T", " ", -1))
```
dszlachta, line 140: I think that you don't have to use `string.Replace` here, just change the format parameter to include `T`: `2006-01-02T15:04:05`

src/apps/chifra/pkg/tslib/tslib.go
```diff
@@ -0,0 +1,143 @@
...
+const tsPath = "/Users/jrush/Development/trueblocks-index/wild.0/unchained/ts.bin"
+
+type Timestamp struct {
+	Bn uint32 `json:"bn"`
+	Ts uint32 `json:"ts"`
+}
+
+var recordCount uint64 = 0
+
+func nRecords() (uint64, error) {
```
dszlachta, line 23: The rest of our codebase uses comments, so that we can generate docs in the future. Should we add comments with descriptions for each of the functions in this file?

dszlachta (commented):
There's a hardcoded path that we may want to remove before merging

src/apps/chifra/pkg/tslib/tslib.go
tjayrush: I want to allow both date strings with spaces and date strings with 'T' to pass.

src/apps/chifra/pkg/tslib/tslib.go
tjayrush: You're right. I'll do that.

src/apps/chifra/pkg/tslib/tslib.go
tjayrush: Of course, you're right here. I'll fix this.

src/apps/chifra/pkg/tslib/tslib.go
tjayrush: Fixed.

src/apps/chifra/pkg/tslib/tslib.go
tjayrush: Actually, on second thought, the name `bn` appears in many, many places through the C++ code, the TypeScript code, and the Go code. I think we will very quickly come to understand exactly what this is. `blockNum` is fine, and also appears in many, many places, so I think we will learn to use both.

src/apps/chifra/pkg/tslib/tslib.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	fileStat, err := os.Stat(tsPath)
+	if err != nil {
+		return 0, err
+	}
+	recordCount = uint64(fileStat.Size()) / 8
+	return recordCount, nil
+}
+
+var loaded = false
```
tjayrush, line 38: @dszlachta

We should discuss this (in these conversations, I suppose).

What do you think about this 'global' memory?

It makes the command-line code very significantly faster if we're converting 100,000s of timestamps in a single invocation of this function (which happens), but it seems a bit precarious in the server code.

I feel like this "issue" will keep coming up (I see it in the `--neighbor` option to `chifra export` as well) and in the names database, so a "pattern" now is better than a pattern later.

dszlachta: I don't see any problem with a global cache other that filling it with values. This would be problematic in the server, because each goroutine could try to initialize the cache.

I think that we could use something like `sync.Cond` to make the goroutines wait until the cache is populated with values: https://stackoverflow.com/a/42772799
What do you think?

tjayrush: I'll merge this and make an issue so we don't forget.

tjayrush:
Merging even though testing is broken.

PR #2012, opened 2022-01-15
Feature nix shell support 

hujw77:
* Build chifra using [nix shell](https://nixos.wiki/wiki/Development_environment_with_nix-shell)
```sh
git clone -b develop https://github.com/trueblocks/trueblocks-core
cd trueblocks-core
nix-shell
mkdir build && cd build
cmake ../src
make
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ If you need help with a step, see the [installation's troubleshooting section](h
     - &#9745; [Install the latest version of Go](https://golang.org/doc/install).
     - &#9745; Install the other dependencies with your command line: `build-essential` `git` `cmake` `python` `python-dev` `libcurl3-dev` `clang-format` `jq`.
 
+Alternatively, for nix users, you can drop into an isolated environment with necessary dependencies with `nix-shell`.
+
 1. Compile from the codebase
     ```shell
     git clone -b develop https://github.com/trueblocks/trueblocks-core
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,12 @@
+with import <nixpkgs> {};
+mkShell {
+    buildInputs = [
+      go
+      git
+      cmake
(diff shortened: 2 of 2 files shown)


tjayrush:
Thanks for this. I'm happy to merge this, as it seems clear, but...

Can you suggest how (and where) to change the installation instructions to include this as an option?

Also, I'm trying hard to keep the higher-up folders as clean as possible, can you move this down into the folder `./src/other/build_assets` which is where we store build-related files?

hujw77:
> Can you suggest how (and where) to change the installation instructions to include this as an option?

https://github.com/TrueBlocks/trueblocks-core/pull/2012/commits/f1c0e293689a583241e31dc41c97466f9df4d1d8

> Also, I'm trying hard to keep the higher-up folders as clean as possible, can you move this down into the folder ./src/other/build_assets which is where we store build-related files?

`shell.nix` in the root directory is a convention. If put the `shell.nix` file to a subdirectory, `nix-shell ./src/other/build_assets/shell.nix` is not friendly.

tjayrush:
> > Can you suggest how (and where) to change the installation instructions to include this as an option?
> 
> [f1c0e29](https://github.com/TrueBlocks/trueblocks-core/commit/f1c0e293689a583241e31dc41c97466f9df4d1d8)
> 
> > Also, I'm trying hard to keep the higher-up folders as clean as possible, can you move this down into the folder ./src/other/build_assets which is where we store build-related files?
> 
> `shell.nix` in the root directory is a convention. If put the `shell.nix` file to a subdirectory, `nix-shell ./src/other/build_assets/shell.nix` is not friendly.

Got it. So people just know if they see `shell.nix` to use it if they want. Sounds good. Running tests now. Will merge if they pass.

PR #2015, opened 2022-01-17
Multi chain 6

tjayrush:


src/apps/chifra/pkg/config/config_paths.go
```diff
@@ -51,7 +51,7 @@ func GetPathToRootConfig() string {
 }
 
 // GetPathToIndex returns the one and only cachePath
-func GetPathToIndex() string {
+func GetPathToIndex(temporarilyUnused string) string {
```
dszlachta, line 54: If it's unused, is there linter error in your editor? If so, try to name the variable `_`, which means "this one is unused"

dszlachta (approved):

PR #2028, opened 2022-01-25
Unchained Index multichain contract

dszlachta:
This PR adds:

- multi-chain version of Unchained Index smart contract
- scripts to generate ABI and golang package to call/deploy the contract
- golang unit tests


src/apps/chifra/pkg/unchained/UnchainedIndex.multichain.sol
```diff
@@ -0,0 +1,60 @@
+/*
+
+See the other file also -- UnchainedIndexv2.sol
+
+Version 2
```
tjayrush, line 5: In a future version 3.0, we might want to consider this post related to the gas cost of publishing the hash as a string. https://ethereum.stackexchange.com/questions/61100/how-much-does-it-cost-to-store-each-ipfs-hash-in-ethereum-blockchain/83025. I guess there are cheaper ways gas-wise. But we won't do anything here. I'll make an issue.

src/apps/chifra/pkg/unchained/unchained/unchainedindex.multichain_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind/backends"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// simulateClient setups a simulated blockchain and deploys the contract
+func simulateClient() (*backends.SimulatedBackend, *bind.TransactOpts, *Unchained) {
```
tjayrush, line 22: This looks super useful...

PR #2096, opened 2022-03-28
Changed IsBlockNumber so it returns the number. Added IsBlockNumberList

### LINKED ISSUE #2094: Need to be able to convert a []string of block ids to actual block numbers
Currently, in our GoLang code, we have routines that validate an array of strings that represent block identifiers (numbers, hashes, and/or special block names such as `latest` or `first`). Other routines validate transaction identifiers.

I need a routine that does that same validation but additionally returns a `[]uint32` array of the 'resolved' block numbers.

**Edit:** On second thought, make a new type called `blockNum` and return `[]blockNum`.

This can be a new routine that calls the existing validation code and then converts the items, or (preferred), a modification to the existing routines to return the converted block numbers. The second of these (i.e., converting and validating in the same routine) avoids a two-pass algorithm. (These lists of blocks can get quite large (100,000's of items), so speed matters.)

The code can be put into a PR against the `develop` branch.

dszlachta:
Closes #2094

### DIFF
--- a/src/apps/chifra/pkg/validate/blockvalidate.go
+++ b/src/apps/chifra/pkg/validate/blockvalidate.go
@@ -32,7 +32,9 @@ func IsBlockHash(str string) bool {
 	return true
 }
 
-func IsBlockNumber(str string) bool {
+type blockNum = uint32
+
+func IsBlockNumber(str string) (bool, blockNum) {
 	base := 10
 	source := str
 
@@ -41,9 +43,27 @@ func IsBlockNumber(str string) bool {
 		source = str[2:]
 	}
 
-	_, err := strconv.ParseUint(source, base, 64)
+	value, err := strconv.ParseUint(source, base, 32)
+	if err != nil {
+		return false, 0
+	}
 
-	return err == nil
+	return true, blockNum(value)
+}
+
+func IsBlockNumberList(strs []string) (bool, []blockNum) {
+	result := make([]blockNum, len(strs))
+
+	for index, stringValue := range strs {
+		check, value := IsBlockNumber(stringValue)
+		if !check {
+			return false, nil
+		}
+
+		result[index] = value
+	}
+
+	return true, result
 }
 
 func IsDateTimeString(str string) bool {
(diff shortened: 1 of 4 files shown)


src/apps/chifra/pkg/validate/blockvalidate.go
```diff
@@ -32,7 +32,9 @@ func IsBlockHash(str string) bool {
 	return true
 }
 
-func IsBlockNumber(str string) bool {
+type blockNum = uint32
```
tjayrush, line 35: I've been thinking we should have this so I can replace all occurrences of uint32 with this. Nice. I think it might want to be in a 'types' package, but don't do that now so the merge into my current branch is easier.

src/apps/chifra/pkg/validate/blockvalidate.go
```diff
@@ -77,7 +97,7 @@ func IsRange(chain, str string) (bool, error) {
 
 	if err == nil {
 		if bRange.Start.Special == "latest" {
-			return false, errors.New("Cannot start range with 'latest'")
+			return false, errors.New("cannot start range with 'latest'")
```
tjayrush, line 100: Nice!

src/apps/chifra/pkg/validate/blockvalidate_test.go
```diff
@@ -29,26 +30,79 @@ func TestIsBlockHash(t *testing.T) {
...
+			args: args{
+				strs: []string{"12147043", "12147042", "12147041"},
+			},
+			want:  true,
+			want1: []blockNum{12147043, 12147042, 12147041},
+		},
+		{
+			name: "list of invalid blocknumbers",
+			args: args{
+				strs: []string{"0xfff", "latest", "newest"},
```
tjayrush, line 81: 0xfff and latest should both return valid block numbers

dszlachta: I'm not sure why we don't support hex block numbers (I remember the discussion about it and there's a comment in the code that we don't want to support them, but I'm not sure why), but when it comes to `latest` I think it should return a block number when it's parsed, while this function should return false (because `latest` is a special name)

PR #2101, opened 2022-04-04
Moves chifra list to pure go lang

tjayrush:


tjayrush:
Replaced with update/better-chifra-list

PR #2110, opened 2022-04-25
Add trueblocksql

naterush:
This PR adds trueblocksql. Included are a few primary files of interest:
1. `create.go` creates and fills an SQL database from the index. Note that currently it hardcodes the index path (which you can update to the location of your index), and only loads the first 15 files just for testing.
2. `list.go` just lets you search the db for a specific address to show you how to do it. 
3. `chunkify.go`. This is just a sample of how you can very very easily take the index that you create and split it into chunks, if you wanna distribute it. They aren't optimized for transmission (which they should be, and this would bring the data size way down to close to pre-war levels). 

The `unchunkify.go` file would similar be like 15 lines of code, which would be nice.

# Instructions

`go run create.go` to create the database
`go run list.go` to test it (good test address is: 1776281d0d40108941cb41e604dcafbdc3cd61a1)
`go run chunkify.go` after running create to test splitting into chunks

# On the overhead of an SQL Database

There is about a 4x storage overhead for this database, which would result in the index being about 250 gbs. This might be minority optimizable through some clever storage techniques, but I don't think by that much...

We could make this much smaller for the initial transmission of the chunks (I think overhead would go down to < 2x if we removed all indexes), so I think the cost doesn't need to be borne in the initial sync or anything.

On data size limitations: sqlite supports up to 281TB of data, so I think we don't need to worry about this.


tjayrush (commented):
Starting a review now. Looks like really nice simple code. I have a question or two.

sql/chunkify.go
```diff
@@ -0,0 +1,87 @@
+package main
```
tjayrush, line 1: I sort of see each of these individual files as a command, so I will probably collapse them all under `chifra`. We have a tool called `chifra chunks` which seems like a likely spot. So `chifra chunks --sql chunkify` or something like that...we'll see.

sql/create.go
```diff
@@ -0,0 +1,190 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+// Deletes the existing database in this file
+func deleteDatabase(databasePath string) {
```
tjayrush, line 18: How is this file different than the previous chunkify.go file? They both seem to delete/create and load the database, but this one loads from Pinata?

sql/create.go
```diff
@@ -0,0 +1,190 @@
...
+	_, err = db.Exec(sqlStmt)
+	if err != nil {
+		log.Printf("%q: %s\n", err, sqlStmt)
+		return
+	}
+}
+
+type Appearance struct {
+	address          string
+	blockIndex       int
```
tjayrush, line 59: I purposefully use uint32 here. Half as big, but still plenty big for 4 billion blocks. I've toyed with uint16 for tx id, but I didn't go there. There's a very large number of transactions, and it's worse because a single transaction can appear multiple times. Every byte counts.

sql/create.go
```diff
@@ -0,0 +1,190 @@
...
+	fmt.Println("Creating DB")
+	_, err = db.Exec(sqlStmt)
+	if err != nil {
+		log.Printf("%q: %s\n", err, sqlStmt)
+		return
+	}
+}
+
+type Appearance struct {
+	address          string
```
tjayrush, line 58: I use bytes here instead of a string. Byte are 20 bytes long. A string is 42 bytes long. We use go/ethereum's common.Address type.

sql/create.go
```diff
@@ -0,0 +1,190 @@
...
+		panic(err)
+	}
+	defer f.Close()
+
+	headerByteSlice := make([]byte, 44)
+	_, err = f.Read(headerByteSlice)
+	if err != nil {
+		log.Fatal(err)
+	}
+	nAddresses := binary.LittleEndian.Uint32(headerByteSlice[44-8 : 44-4])
```
tjayrush, line 81: There's a package in chifra/pkg called index that encapsulates the index file as it exists on disk. I'm 99% sure there's a read function there.

sql/create.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	appearanceByteSlice := make([]byte, 8*nAppearances)
+	_, err = f.Read(appearanceByteSlice)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	for i := 0; int64(i) < int64(nAddresses); i++ {
+		// TODO: I am not sure this hex encoding is correct, so check it
```
tjayrush, line 100: We can use the package. That way an optimization there will benefit here, plus it works and is well tested.

sql/create.go
```diff
@@ -0,0 +1,190 @@
...
+			appearances = append(appearances, Appearance{address: address, blockIndex: int(blockIndex), transactionIndex: int(transactionIndex)})
+		}
+	}
+
+	return appearances, nil
+}
+
+// Fills a database with the first X chunks of the index, for testing
+func fillDatabase(databasePath string, indexFolderPath string) {
+	db, err := sql.Open("sqlite3", databasePath)
```
tjayrush, line 118: Couldn't be simpler.

sql/create.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer db.Close()
+
+	tx, err := db.Begin()
+	if err != nil {
+		log.Fatal(err)
+	}
+	addressesInsertStmt, err := tx.Prepare("INSERT OR IGNORE INTO addresses(address) VALUES(?)")
```
tjayrush, line 128: Just for my information, what does OR IGNORE do?

sql/create.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+		}
+		if i == 15 {
+			break
+		}
+	}
+	tx.Commit()
+}
+
+// Deletes existing database, creates a new one, and fills it
```
tjayrush, line 174: Why do you want to delete the existing database? Is that just for testing or something?

sql/unchunkify.go
```diff
@@ -0,0 +1,88 @@
+package main
```
tjayrush, line 1: I don't think this is complete, is it?

tjayrush (commented):
Hey Nate. Thanks for this.

I made some comments in the code and had a few questions. A couple of things I want to figure out before merging:

1. Sharing via an immutable data store (as you know) is a very important part of what we built, so I want to make sure that is preserved. I'm sure it is, because this SQL stuff can be local-only and local-after-download-as-an-option. But, at the same time, it will come in handy when (and if) we go 'centralized'.

2. I want to incorporate it into the regular chifra command line (and server code). The easiest way to do this would be to generate 'stubs' using my existing auto-code-generation stuff and then this will very easy fit into that code with no problem. It will probably fit somewhere in `chifra chunks`.

3. I want to use some of our existing code for the index reading/writing for example. It's well tested and basically the same as this. So, no duplicated code. We can do that once it's merged.

4. I like having the database as part of the build. I have a feeling I will be using it elsewhere (names database for example).

5. Databases are great for servers because you can start up the server, load the database once at initialization and then server data very, very quickly. For commands line tools, it's not so clear. Because of the way our code works -- command line first -- sever simply calls out to the command line on each request -- I'm not sure this will be faster -- I'm interested to learn. Since we're porting everything to go, eventually there won't be this 'to the command line' usage pattern, so then it will be a more fair comparison and in the server case will be a clear win. I'm interested to learn.

tjayrush:
A comment about size. In our existing code, we do three things to make the database smaller. We have two tables, as you do, one has `<address><start><nrows>` and the second has `<blknum><txid>`.

We do a couple of things to make it smaller:

1. We use int32 for all integers (as opposed to int which is an int64 on most machines). Makes the data half as big. Likely a very small performance hit well worth it for 1/2 size.
2. We use bytes instead of a string to store the address. Again, half as big and exactly as fast (if not faster)
3. We do not store an id in either table. We take advantage of the fact that the file is immutable (as a chunk) to eliminate the need for a primary key. The positioning of the rows in the second table are fixed (since we're no longer adding new records). This eliminates an interger in every row of the address table and another one in every row of the appearance table, but it does require chunking so it's immutable.

And one thing that makes it faster:

1. All the records are fixed width, so we can read the entire file in a single read. (Databases do this too, so it's a wash.)

Just some thoughts.

tjayrush:
Two other things please before I merge:

1. Can you move this into `./src/examples/sql`? That's a good place to keep it for now until it gets into its own command.
2. Can you re-base to develop? If you do, I think the auto-tests will pass

tjayrush:
This code was added in the examples folder, so this PR can be closed. Thakns.

PR #2117, opened 2022-05-04
Multichain JSON Manifest

dszlachta:
This is the new, multi-chain, JSON manifest format from the branch that we want to delete

### DIFF
--- a/src/apps/chifra/pkg/pinlib/download_test.go
+++ b/src/apps/chifra/pkg/pinlib/download_test.go
@@ -13,13 +13,16 @@ import (
 
 var manifestJSONSource = `
 {
-  "fileName": "pin-manifest.json",
+  "version": "2",
+  "chain": "mainnet",
+  "chainId": 1,
   "indexFormat": "Qmart6XP9XjL43p72PGR93QKytbK8jWWcMguhFgxATTya2",
   "bloomFormat": "QmNhPk39DUFoEdhUmtGARqiFECUHeghyeryxZM9kyRxzHD",
   "commitHash": "f29699d3281e41cb011ddfbe50b7f01bfe5e3c53",
-  "prevHash": "QmP4i6ihnVrj8Tx7cTFw4aY6ungpaPYxDJEZ7Vg1RSNSdm",
-  "newBlockRange": "000000000-000350277",
-  "newPins": [
+  "names": "QmP4i6ihnVrj8Tx7cTFw4aY6ungpaPYxDJEZ7Vg1RSNSdm",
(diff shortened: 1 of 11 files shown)


src/apps/chifra/pkg/pinlib/download_test.go
```diff
@@ -13,13 +13,16 @@ import (
...
+  "version": "2",
+  "chain": "mainnet",
+  "chainId": 1,
   "indexFormat": "Qmart6XP9XjL43p72PGR93QKytbK8jWWcMguhFgxATTya2",
   "bloomFormat": "QmNhPk39DUFoEdhUmtGARqiFECUHeghyeryxZM9kyRxzHD",
   "commitHash": "f29699d3281e41cb011ddfbe50b7f01bfe5e3c53",
-  "prevHash": "QmP4i6ihnVrj8Tx7cTFw4aY6ungpaPYxDJEZ7Vg1RSNSdm",
-  "newBlockRange": "000000000-000350277",
-  "newPins": [
+  "names": "QmP4i6ihnVrj8Tx7cTFw4aY6ungpaPYxDJEZ7Vg1RSNSdm",
```
tjayrush, line 22: We currently make local changes to this file, but I want to change this so there's a 'cononical' file (this one) and an 'edited' file. We can do that once we move over to All Go Code for names.

In the meantime, since there's no real way to edit this file (other than by hand), we can just download it directly.

There's two files. One is a binary cache of the names.tab file. The other is the names.tab file itself. If names.tab is newer, names.bin gets created.

Upshot: We should store the `.tab` file on IPFS. When we download it (if the blockNumber's timestamp is later than the local file), and touch it. Then the .bin will get updated.

We can discuss.

src/apps/chifra/pkg/pinlib/download_test.go
```diff
@@ -13,13 +13,16 @@ import (
...
+  "timestamps": "QmcvjroTiE95LWeiP8HHq1YA3ysRchLuVx8HLQui8WcSBV",
```
tjayrush, line 23: Same comment here, but in this case, the timestamp file will have been updated, so it's quite likely that a 'normal' user will never download this. There is only a binary file. The first time user, will want this file.

In both cases (names and timestamp) we will want to remove these files from our repo -- including back version if possible since these two files bloat the size of the repo hugely.

tjayrush, line 23: Same comment here, but in this case, the timestamp file will have been updated, so it's quite likely that a 'normal' user will never download this. There is only a binary file. The first time user, will want this file.

In both cases (names and timestamp) we will want to remove these files from our repo -- including back version if possible since these two files bloat the size of the repo hugely.

PR #2140, opened 2022-05-30
DO NOT MERGE: Wrapping global data access with a mutex

tjayrush:


tjayrush:
@dszlachta Can you take a look at this very simple PR?

It's clear to me that the access to the global `var` is not okay without a mutex, but it doesn't fix a problem I'm having with the scraper.

When I run the scraper without this mutex, I get the same error described here: https://stackoverflow.com/questions/45585589/golang-fatal-error-concurrent-map-read-and-map-write.

But when I make the change in this PR, expecting it to work, it doesn't.

I can fix the issue by removing the global variable entirely and just doing the Dial each time this function is called, but the function is call A LOT and it slows things down noticeably.

Any advice?

dszlachta:
@tjayrush it doesn't work, because you're creating a new mutex everytime `GetClient` is called, so this new mutex is never locked when the code tries to lock it (and because of that the mutex doesn't block and you get the error). I think it's safe to make this mutex a global variable in the package, provided that you won't export it. Mutexes are concurrency-safe and if `GetClient` is the only function to use, it should be fine.

tjayrush:
Makes perfect sense. Thanks.

PR #2157, opened 2022-06-07
Finally fixing scraper

tjayrush:


src/apps/chifra/pkg/pinlib/manifest/json_manifest.go
```diff
@@ -10,11 +10,12 @@ import (
 )
 
 // ReadJSONManifest reads manifest in JSON format
-func ReadJSONManifest(reader io.Reader) (*Manifest, error) {
+func ReadJSONManifest(chain string, reader io.Reader) (*Manifest, error) {
 	decoder := json.NewDecoder(reader)
 	manifest := &Manifest{}
 
 	err := decoder.Decode(manifest)
+	manifest.Chain = chain
```
dszlachta: Shouldn't this be in the manifest file? So the manifest is the source of truth when it comes to to which chain it belongs?

PR #2166, opened 2022-06-10
Workaround for monitors issue

dszlachta:
For each monitor being freshened, we lock a mutex. This makes sure we won't try to move a monitor that has already been moved.


tjayrush:
Answer one quick question please, then I'll merge this.

Issue #2164, opened 2022-06-10
chifra export poor performance extracting statements


artur-jablonski:
`time chifra export --fmt json --accounting --articulate 0x6e4c6D9B0930073e958ABd2ABa516b885260b8Ff`

First time export is very slow:

```
Executed in  107,05 mins   fish           external
   usr time   32,83 mins  678,00 micros   32,83 mins
   sys time   47,20 mins   77,00 micros   47,20 mins
```

At the time of running it, the index reported `35571` transactions, which is not that many (TM).

This is first time export (no cache) against local erigon node

Most of the time is spent in reconciliation judging by the output while this is running.

Can anyone run this for me and time it?

nazt:
3205698287 (   7437)- <PROG>  : Completed      35492 of    35492 txs for address 0x6e4c6d9b0930073e958abd2aba516b885260b8ff

________________________________________________________
```
Executed in   67.06 mins   fish           external
   usr time   16.46 mins  1094.00 micros   16.46 mins
   sys time   37.42 mins    0.00 micros   37.42 mins
```

artur-jablonski:
> 3205698287 ( 7437)- : Completed 35492 of 35492 txs for address 0x6e4c6d9b0930073e958abd2aba516b885260b8ff
> 
> ```
> Executed in   67.06 mins   fish           external
>    usr time   16.46 mins  1094.00 micros   16.46 mins
>    sys time   37.42 mins    0.00 micros   37.42 mins
> ```

Thx!. So you're a bit faster, but still far from fast.... My SSD is connected via USB-C so it could explain the difference I suppose if your drive is plugged in directly to your mainboard. I mean this is speculation but I am assuming such setup would give faster IO.... is this how you have it setup out of curiousity?

EDIT: or it could be that you didn't blow away your cache. I removed everything from cache folder just to see what's the performance when there's no cache involved.

EDIT2: also it could be that you have more solid connection for pulling ABIs from etherscan since this run involves articulation

artur-jablonski:
One inefficiency I found in the code is this:

https://github.com/TrueBlocks/trueblocks-core/blob/c0fbc744d51e6a6935f133f89ec50f800af27f5c/src/apps/acctExport/handle_accounting.cpp#L201-L211

[...]

Issue #2162, opened 2022-06-09
chifra export intrablock transactions for same ERC20 tokens and address are collapsed into one transaction


artur-jablonski:
When you are extracting statements for an address that inside same block has multiple transactions that move the same ERC20 token, then the overall net value of all transfers in the block for that token will be associated with the first transaction in the block. The other transactions will not be shown at all.

**For example:**
In block `14933657` the address: `0x6aED588ca2052CCFC907Db8c24dF4B7B95A29A5E`
peforms 2 transfers of YFI token (`0x0bc529c00C6401aEF6D220BE8C6Ea1667F6Ad93e`) to two different recipients:

1. The first one in the block:
txIndex: 303
txHash: `0x458ae31579274588f04850853ec71f67e1e21b2d4e72ada18fdc04352fc22295`
to: `0xACA096d2E6492D7380c5d8E70D90E13EEcC72dd4`
value: `304442430000000000`

2. The second one in the block:
txIndex: 305
txHash: `0x8c1703f5dee17242784c912f208f55508cf003f2b3b0ca9070d9556a619df186`
to: `0xcc34cc875d2d052f5b1c19652b5e71d6f90c7f7d`
value: `961090000000000`

When calling:
`chifra export --fmt json --accounting --articulate 0x6aED588ca2052CCFC907Db8c24dF4B7B95A29A5E`

You can see that under the first transaction in block (with index 303) the following statement is generated for the YFI token:

```
{
      "blockNumber": 14933657,
      "transactionIndex": 303,
      "timestamp": 1654793734,
      "assetAddr": "0x0bc529c00c6401aef6d220be8c6ea1667f6ad93e",
      "assetSymbol": "YFI",
      "decimals": 18,
      "prevBlk": 14927653,
      "prevBlkBal": "18255476410000000000",
      "begBal": "18255476410000000000",
      "endBal": "17950072890000000000",
      "amountIn": "",
      "internalIn": "",
      "selfDestructIn": "",
      "minerBaseRewardIn": "",
      "minerNephewRewardIn": "",
      "minerTxFeeIn": "",
      "minerUncleRewardIn": "",
      "prefundIn": "",
      "amountOut": "305403520000000000",
      "internalOut": "",
      "selfDestructOut": "",
      "gasCostOut": "",
      "spotPrice": 7616.14451,
      "priceSource": "uniswap",
      "begBalDiff": "",
      "endBalCalc": "17950072890000000000",
[...]

PR #2167, opened 2022-06-10
changing condition of adding erc20 token candidates

artur-jablonski:


tjayrush:
Thanks for this PR. This one-liner was a bit too specific for a single topic, so I extended it a bit to be more general. I added it with a test case to this PR https://github.com/TrueBlocks/trueblocks-core/pull/2173 (which has an additional test). I'll close this PR. Thanks for it.

artur-jablonski:
> Thanks for this PR. This one-liner was a bit too specific for a single topic, so I extended it a bit to be more general. I added it with a test case to this PR #2173 (which has an additional test). I'll close this PR. Thanks for it.

I scanned through the changes. My understanding is that the "statements" in trueblocks are all about transfer of value. If there's no transfer of value there's no statement. Correct? The only event that indicates there was any transfer of value of ERC20 from the spec (https://ethereum.org/en/developers/docs/standards/tokens/erc-20/) is the Transfer event:

```
event Transfer(address indexed _from, address indexed _to, uint256 _value)
event Approval(address indexed _owner, address indexed _spender, uint256 _value)
```

so I thought this should be the only event to look for. `Approval` is not transfering value nor do (I assume) other events from `tokenRelated` array that I am not sure exactly what they are (which standard they come from):

```
make_pair("0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
              true),  // Transfer(address from, address to, uint256 value)
    make_pair("0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925",
              true),  // Approval(address owner, address spender, uint256 value)
    make_pair("0xd4735d920b0f87494915f556dd9b54c8f309026070caea5c737245152564d266",
              true),  // Transfer(bytes32 node, address owner)
    make_pair("0x30385c845b448a36257a6a1716e6ad2e1bc2cbe333cde1e69fe849ad6511adfe",
              true),  // Minted(address,uint256)
};
```

Perhaps I am missin gsth

tjayrush:
>  If there's no transfer of value there's no statement. Correct?

Not quite.

You're right that many of these `isTokenRelated` topics and four-bytes are not standard.

But I'm trying to reconcile everything, so standards, while super nice, aren't enough.

I've not seen a transfer of value during an Approval, but it's possible.

I have seen it in the other things included in `isTokenRelated`.

My thought is to be 'over-inclusive' and if we "reconcile" a transaction that has no value transfer that's better than not reconciling one that does.

I do this frequently -- especially in the indexing and bloom filtering on the assumption it's easier to filter out data later in the pipeline than to never have had the data to being with.

artur-jablonski:
> > If there's no transfer of value there's no statement. Correct?
> 
> Not quite.
> 
> You're right that many of these `isTokenRelated` topics and four-bytes are not standard.
> 
> But I'm trying to reconcile everything, so standards, while super nice, aren't enough.
> 
> I've not seen a transfer of value during an Approval, but it's possible.
> 
> I have seen it in the other things included in `isTokenRelated`.
> 
> My thought is to be 'over-inclusive' and if we "reconcile" a transaction that has no value transfer that's better than not reconciling one that does.
> 
> I do this frequently -- especially in the indexing and bloom filtering on the assumption it's easier to filter out data later in the pipeline than to never have had the data to being with.

Got ya. 

On the flip side being overinclusive can result in poorer performance but I think I understand where you're coming from with this approach. In any case I think the logic as it is now is way better than before so it's a win :muscle:

tjayrush:
Once we get to GoLang, we can optimize, but by then, I hope, we have solid tests (especially surrounding reconciliation). Tests make optimizations much, much easier so we're not worried about breaking things.

On that note, I tried to add your test to the regular testing, but something was broken. One thing that would be good, while we're waiting for the GoLang code, would be to enhance the test cases by finding non-reconciling examples. (For test cases, it's always better to find small accounts with errors -- otherwise, the testing gets super slow.)

tjayrush:
Thanks so much for your thoughts, Artur. I appreciate them a lot.

artur-jablonski:
> On that note, I tried to add your test to the regular testing, but something was broken.

Hmmm, I think this PR didn't fix anything in reconciliation logic but rather improved performance by not including all log entries as tokens when just one of them was token related (which was the logic before my PR), so there are no tests to add/remove/adjust. Maybe you're referring to the other PR I opened for partial (intra-block) reconciliation of ERC20? (#2162)

tjayrush:
No. This one. I tried to add the test, but it was failing, so I had to turn it off (making it say `local` turns it off by the way).

Help me simplify this a bit (if you’re not burned out…)


Send me the exact command you use and the exact respected results. Hopefully, it doesn’t work on a release prior to the fix, but it does after the fix.

Sorry if you’ve already done that, I’ve lost the thread a bit….

I will

1. run it against an old release to make sure it fails
2. move to the latest release
3. work on it until the test passes
4. Push new code if I need to.

I’d really like to capture a non-reconciling token transaction if I can.

Of course, I understand if I’m asking too much, and thanks so much for all your help so far.

Cheers.

artur-jablonski:
So there was no non-reconciling token in this case. I found the issue by reading the code. I explained it here: https://github.com/TrueBlocks/trueblocks-core/issues/2164#issuecomment-1152442993.

So I would expect that this PR would simply not break any tests. If it does, then I can have a look why. But esentially this is an optimisation.

PR #2183, opened 2022-06-22
Serve API error handling

Skemba:
Added fatal log on RPC loading failed.


Skemba:
@tjayrush let me know if there's a reason for server to not exit if RPC loading failed and I'll change Fatalf to Printf.

tjayrush:
> @tjayrush let me know if there's a reason for server to not exit if RPC loading failed and I'll change Fatalf to Printf.

I think I'd prefer that the server doesn't die, but it should also probably return an error response (which it may not currently do). If a front end is running against the server, and it dies, that would not be good.

Skemba:
> If a front end is running against the server, and it dies, that would not be 

This runs only once when `chifra serve` is executed, so it would die right away, someone who started it would know right away. which might actually be the best behavior of a server, rather than it saying it's "running" while failing?

tjayrush:
Sounds good. I’ll merge it now.

tjayrush:
Thanks for this. Sorry it took so long to merge.

PR #2195, opened 2022-07-06
Removed code that was breaking DAppNode

dszlachta:
Prefixing ports with `http://localhost` breaks DAppNode, because the package has to use `0.0.0.0` as host. This code was causing the server to listen on `http://localhost0.0.0.0:8080`. According to Go documentation, `localhost` is the default value if the host is missing, so this code was redundant.

### DIFF
--- a/src/apps/chifra/server/run.go
+++ b/src/apps/chifra/server/run.go
@@ -6,7 +6,6 @@ package servePkg
 
 import (
 	"log"
-	"strings"
 
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/colors"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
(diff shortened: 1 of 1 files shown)


src/apps/chifra/server/run.go
```diff
@@ -24,9 +23,6 @@ func RunServe(cmd *cobra.Command, args []string) error {
 	}
 
 	apiUrl := opts.Port
-	if !strings.Contains(apiUrl, "http") {
-		apiUrl = "http://localhost" + apiUrl
-	}
```
tjayrush, line 26: This value is only used for as a display item below. Is that correct? If yes, does having this code break anything or is it just confusing? Can you check to see if the string starts with ` colon` (that is, it's only a port) and then add this string it does? In your case, the opts.Port must be `0.0.0.0:xxx`.

Another option would be to separate host from port on the command line, and if the host is empty, it could insert the default.

Your thoughts?

dszlachta: You are right, I was rushing with it and got a bit confused. The issue was somewhere else, in Docker setup, so that's why I thought that this change fixes something :) We can ignore it or merge to get rid of printing things like `http://localhost0.0.0.0:8080`

PR #2211, opened 2022-07-31
export: remove first_record and max_records block checks

snf:
first_record and max_records are not blocks and can be larger than the latest block numbers.
An example where these checks fail is trying to get the latest N records of a very active address' appearances


tjayrush:
This is an easy change, but it's actually part of auto generated code. (Auto generated code feeds into the API server, the TypeScript SDK, the command line processing, and the automatically generated documentation.)

I can't merge this PR, but I will make this change in a different PR that I will both link here and merge into the develop branch.

Thanks for finding this. Once it's merged, it would be great if you can test it. I'll do this now.

snf:
Awesome, thank you! I suspected that those `<...>` types must had some usecase. I'll test during the weekend but I don't see why it wouldn't work

PR #2212, opened 2022-08-02
chifra server: fixing race condition

snf:
I was getting a `panic: send on closed channel` on this channel and tracing the code back, it looks like the code [here](https://github.com/TrueBlocks/trueblocks-core/blob/master/src/apps/chifra/internal/list/handle_freshen.go#L153) is waiting on the signal to close the channel. I believe that generates a race condition because the `Done` signal is sent before the data could be sent to the channel. If the thread preempts in between and the other one closes the channel, then the data cannot be sent and it panics.

This is only a theory but I'm triggering the issue consistenly after using the api server for around an hour.


snf:
Update. Ran the chifra server for a day with this patch and it hasn't crashed while it was crashing every hour without it.

tjayrush:
Wow. So excellent. @dszlachta Can this looks super inoccuous. Does it make sense to you? I know you didn't write this code, but you're better at Go than I am.

@snf Can you change this to merge to develop?

PR #2257, opened 2022-08-31
Adding links to buttons (Twitter…)

omahs:
Adding links to buttons (Twitter, Forks…)


tjayrush:
Don't they already have links? I'm confused? You have square brackets inside of square backets, not?

omahs:
Hi @tjayrush, at the moment those are just links to the button image, with the change it will go to the twitter page.

tjayrush:
Are you in our discord? I think you get a POAP. It would be cool (if you care) to show the rest of the visitors to our discord how exciting it is to get POAPs -- plus it would be interesting to see how it works.

omahs:
I'm not on the Discord but I'll join and explain a little about the POAP process 😃

PR #2260, opened 2022-09-01
Output streaming

dszlachta:
This PR adds output.Stream* functions, models and type for Receipt/RawReceipt

### DIFF
--- a/src/apps/chifra/internal/globals/output.go
+++ b/src/apps/chifra/internal/globals/output.go
@@ -91,11 +91,22 @@ func (opts *GlobalOptions) RenderObject(data interface{}, first bool) error {
 	if opts.Writer == nil {
 		log.Fatal("opts.Writer is nil")
 	}
-	return output.OutputObject(data, opts.Writer, opts.Format, opts.NoHeader, opts.ApiMode, first, nil)
+
+	format := opts.Format
+	if opts.Raw {
+		// If users wants raw output, we will most probably print JSON
+		format = "json"
+	}
+	return output.OutputObject(data, opts.Writer, format, opts.NoHeader, opts.ApiMode, first, nil)
 }
 
 // TODO: Fix export without arrays
 func (opts *GlobalOptions) RenderHeader(data interface{}, w *io.Writer, format string, apiMode, hideHeader, first bool) error {
+	if opts.Raw {
+		// We don't render the header if user wants raw output.
+		return nil
+	}
+
 	if apiMode {
 		// We could check this, but if it's not empty, we know it's type
 		hw, _ := (*w).(http.ResponseWriter)
@@ -118,6 +129,11 @@ func (opts *GlobalOptions) RenderHeader(data interface{}, w *io.Writer, format s
 
 // TODO: Fix export without arrays
 func (opts *GlobalOptions) RenderFooter() error {
+	if opts.Raw {
+		// We don't render the footer if user wants raw output.
+		return nil
+	}
+
 	if opts.Format == "api" || opts.Format == "json" {
 		opts.Writer.Write([]byte("\n  ]"))
 		showMeta := opts.ApiMode || opts.Format == "api"
@@ -133,7 +149,7 @@ func (opts *GlobalOptions) RenderFooter() error {
 			}
 			opts.Writer.Write(b)
 		}
-		opts.Writer.Write([]byte("\n}"))
+		opts.Writer.Write([]byte("\n}\n"))
 	}
 	return nil
 }
--- a/src/apps/chifra/internal/receipts/output.go
+++ b/src/apps/chifra/internal/receipts/output.go
@@ -10,8 +10,12 @@ package receiptsPkg
 
 // EXISTING_CODE
 import (
+	"fmt"
 	"net/http"
 
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/output"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/rpcClient"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
 	"github.com/spf13/cobra"
 )
 
@@ -45,12 +49,43 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
 	if opts.Globals.ApiMode {
 		return nil, false
 	}
+	if opts.Articulate {
+		err = opts.Globals.PassItOn("getReceipts", opts.Globals.Chain, opts.toCmdLine(), opts.getEnvStr())
+		return err, true
+	}
 
-	handled = true
-	err = opts.Globals.PassItOn("getReceipts", opts.Globals.Chain, opts.toCmdLine(), opts.getEnvStr())
-	// EXISTING_CODE
+	getTransaction := func(models chan types.Modeler[types.RawReceipt], errors chan error) {
+		for idIndex, rng := range opts.TransactionIds {
+			txList, err := rng.ResolveTxs(opts.Globals.Chain)
+			if err != nil {
+				errors <- err
+				return
+			}
+			for _, tx := range txList {
+				receipt, err := rpcClient.GetTransactionReceipt(opts.Globals.Chain, uint64(tx.BlockNumber), uint64(tx.TransactionIndex))
+				if err != nil && err.Error() == "not found" {
+					errors <- fmt.Errorf("transaction %s not found", opts.Transactions[idIndex])
+					return
+				}
+				if err != nil {
+					errors <- err
+					return
+				}
 
-	return
+				models <- &receipt
+			}
+		}
+	}
+	err = output.StreamMany(opts.Globals.Writer, getTransaction, output.OutputOptions{
+		ShowKeys:   !opts.Globals.NoHeader,
+		ShowRaw:    opts.Globals.Raw,
+		ShowHidden: opts.Globals.Verbose,
+		Format:     opts.Globals.Format,
+	})
+	if err != nil {
+		return err, true
+	}
+	return nil, true
 }
 
 // EXISTING_CODE
--- a/src/apps/chifra/internal/receipts/output.go
+++ b/src/apps/chifra/internal/receipts/output.go
@@ -18,10 +18,12 @@ import (
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/output"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/rpcClient"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/utils"
 	"github.com/spf13/cobra"
 )
 
 // EXISTING_CODE
+var byzantiumBlockNumber = 4370000
 
 // RunReceipts handles the receipts command for the command line. Returns error only as per cobra.
 func RunReceipts(cmd *cobra.Command, args []string) (err error) {
@@ -38,6 +40,18 @@ func ServeReceipts(w http.ResponseWriter, r *http.Request) (err error, handled b
 	// EXISTING_CODE
 	// EXISTING_CODE
 	return opts.ReceiptsInternal()
+}
+
+// TODO: remove this function when rewrite to Go is completed. It is only used to send
+// pre-Byzantium transactions to C++ version
+func getReceiptsCmdLine(opts *ReceiptsOptions, txs []string) string {
+	options := ""
+	if opts.Articulate {
+		options += " --articulate"
+	}
+
+	options += " " + strings.Join(txs, " ")
+	return options
 }
 
 // ReceiptsInternal handles the internal workings of the receipts command.  Returns error and a bool if handled
@@ -62,6 +76,11 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
 			fmt.Fprintln(os.Stderr, err)
 		}
 	}()
+	clientVersion, err := rpcClient.GetVersion(opts.Globals.Chain)
+	if err != nil {
+		return err, true
+	}
+	erigonUsed := utils.IsClientErigon(clientVersion)
 
 	getTransaction := func(models chan types.Modeler[types.RawReceipt], errors chan error) {
 		for idIndex, rng := range opts.TransactionIds {
@@ -76,6 +95,15 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
 				return
 			}
 			for _, tx := range txList {
+				if tx.BlockNumber < uint32(byzantiumBlockNumber) && !erigonUsed {
+					err = opts.Globals.PassItOn("getReceipts", opts.Globals.Chain, getReceiptsCmdLine(opts, []string{rng.Orig}), opts.getEnvStr())
+					if err != nil {
+						errors <- err
+						return
+					}
+					continue
+				}
+
 				receipt, err := rpcClient.GetTransactionReceipt(opts.Globals.Chain, uint64(tx.BlockNumber), uint64(tx.TransactionIndex))
 				if err != nil && err.Error() == "not found" {
 					notFound = append(notFound, fmt.Errorf("transaction %s not found", opts.Transactions[idIndex]))
--- a/src/apps/chifra/internal/receipts/output.go
+++ b/src/apps/chifra/internal/receipts/output.go
@@ -12,6 +12,8 @@ package receiptsPkg
 import (
 	"fmt"
 	"net/http"
+	"os"
+	"strings"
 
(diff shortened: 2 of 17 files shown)


tjayrush:
Good start. Two things just to explain something:

1) The `RenderHeader`, `RenderFooter`, `RenderObject` stuff allows us to "stream" and produce various formats (`json`, `txt`, `csv` -- and others in the future). In `txt` and `csv` mode, it prints the headers (if not turned off), and in `json` mode, it opens the object. The `RenderFooter` closes the object in `json` mode. I envision at least two additional formats: `table` (or `pdf`) that would print something like a PFD bank statement or an HTML table, and `lambda` or similar to shove the data into an endpoint. In the cases where we can't stream for some reason (or choose not to), you can use `RenderSlice`.

2) All the Simple types in the `types` package allow me to have an `internal` (generally hidden from the rest of the code) and an `external` representation for a given data structure. Many times, there are fields in the raw data that I don't want to show or (more usual) there are fields such as summaries that I need that don't belong on the raw data. Also, these data types can be automatically generated and therefore well documented in the future. They may disappear in the future, but I don't want to lose the ability to auto-generate docs.

src/apps/chifra/internal/globals/output.go
```diff
@@ -131,7 +147,7 @@ func (opts *GlobalOptions) RenderFooter() error {
 			b, err := json.MarshalIndent(meta, "  ", "  ")
 			opts.Writer.Write(b)
 		}
-		opts.Writer.Write([]byte("\n}"))
+		opts.Writer.Write([]byte("\n}\n"))
```
tjayrush, line 152: Let's put a TODO list item here. I think that we can "correct the test cases" as opposed to do this only to make the test cases work. I'm not sure if the right solution is "no" extra line feed or "yes" extra line feed, but let's correct the gold test cases as opposed to generating less "canonical" data. But don't let this delay your work -- thus the TODO maker.

src/apps/chifra/internal/receipts/output.go
```diff
@@ -46,42 +46,53 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
...
+
+	// TODO: support options
+	// err = opts.Globals.RenderHeader(h.Model(opts.Globals.Verbose).Data, &opts.Globals.Writer, opts.Globals.Format, opts.Globals.ApiMode, opts.Globals.NoHeader, true)
+	if err != nil {
+		return err, true
+	}
+
+	cnt := 0
+	for idIndex, rng := range opts.TransactionIds {
+		txList, err := rng.ResolveTxs(opts.Globals.Chain)
```
tjayrush: Another TODO list item (in other words, don't fix it now, but note it). I've been thinking that what we should do here instead of `Resolving` (which builds an array of transactions ids to traverse), we should stream through this list and do the traversal on the fly. An example of where this breaks down is the block range (which doesn't apply here, but does for `chifra blocks`, for example) of `chifra blocks 0-latest:2` which generates an array with 8,000,000 records. This should be "streamed" as well.

src/apps/chifra/internal/globals/output.go
dszlachta: Sure. One reason to have the newline is printing JSON on command line. Without new line, `}` appears just before the shell prompt (on the same line). It's little frustrating :P

src/apps/chifra/internal/receipts/output.go
dszlachta: Great point and this is something I've been working on, I just don't have the code yet.

src/apps/chifra/internal/globals/output.go
tjayrush: Sounds good.

src/apps/chifra/internal/receipts/output.go
tjayrush: I feel like this might be a bit more complicated because it touches code that is used by a lot of code. Let's think of this as a "nice to have" and not open this particular can of worms too wide just yet. Unless it's not as complicated as I think it is.

src/apps/chifra/internal/receipts/output.go
```diff
@@ -45,12 +65,77 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
 	if opts.Globals.ApiMode {
```
tjayrush, line 52: We should be able to remove this test and allow ApiMode to go straight through the same as command line.

src/apps/chifra/internal/receipts/output.go
```diff
@@ -10,12 +10,20 @@ package receiptsPkg
...
 
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/output"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/rpcClient"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/utils"
 	"github.com/spf13/cobra"
 )
 
 // EXISTING_CODE
+var byzantiumBlockNumber = 4370000
```
tjayrush: At the bottom of this file is are two comments called `// EXISTING_CODE`. Put any code you want to preserve inside of those comments. The auto-code generator only preserves things inside of an `// EXISTING_CODE` pair. Same comment for `getReceiptsCmdLine` below.

src/apps/chifra/internal/receipts/output.go
```diff
@@ -34,6 +42,18 @@ func ServeReceipts(w http.ResponseWriter, r *http.Request) (err error, handled b
 	return opts.ReceiptsInternal()
 }
 
+// TODO: remove this function when rewrite to Go is completed. It is only used to send
+// pre-Byzantium transactions to C++ version
+func getReceiptsCmdLine(opts *ReceiptsOptions, txs []string) string {
```
tjayrush: Put this at the bottom of the file inside of `// EXISTING_CODE` comment pair.

tjayrush: Is there a reason we cannot use this routine: https://github.com/TrueBlocks/trueblocks-core/blob/master/src/apps/chifra/internal/receipts/options.go#L55?

src/apps/chifra/internal/receipts/output.go
```diff
@@ -45,12 +65,77 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
...
+		return err, true
+	}
 
-	handled = true
-	err = opts.Globals.PassItOn("getReceipts", opts.Globals.Chain, opts.toCmdLine(), opts.getEnvStr())
-	// EXISTING_CODE
+	notFound := make([]error, 0)
+	defer func() {
+		for _, err := range notFound {
+			fmt.Fprintln(os.Stderr, err)
```
tjayrush, line 63: This should be reported through standard err (using `logger.Error`) so it doesn't intermix with the data produced. 

The question to ask is this: if a command-line user pipes the output into a file, should the error message be in the file or should it be reported to the end user?

src/apps/chifra/internal/receipts/output.go
```diff
@@ -45,12 +65,77 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
...
+	}()
+	clientVersion, err := rpcClient.GetVersion(opts.Globals.Chain)
+	if err != nil {
+		return err, true
+	}
+	erigonUsed := utils.IsClientErigon(clientVersion)
 
-	return
+	getTransaction := func(models chan types.Modeler[types.RawReceipt], errors chan error) {
+		for idIndex, rng := range opts.TransactionIds {
```
tjayrush, line 74: Add a TODO item above this line that says "// TODO: stream transaction identifiers

src/apps/chifra/pkg/output/stream.go
```diff
@@ -0,0 +1,177 @@
...
+	tmpl, err := template.New("").Parse(options.Format)
+	if customFormat && err != nil {
+		return err
+	}
+
+	for {
+		select {
+		case model, ok := <-models:
+			if !ok {
+				continue
```
tjayrush, line 146: Will this ever happen? If not, can we panic here instead of silently ignore?

src/apps/chifra/pkg/rpcClient/client.go
```diff
@@ -99,6 +100,29 @@ func GetIDs(provider string) (uint64, uint64, error) {
 	return ch.Uint64(), n.Uint64(), nil
 }
 
+// TODO: C++ code used to cache version info
```
tjayrush, line 103: We tried to cache everything we ever asked of the node. The node is SLOW. This brings up an interesting issue. There's a global option called `--cache` which, if one, should cache the returned value from the node. If that returned value is present, we should read it. Super simple cache since we can safely assume the data is immutable (if it's old enough -- six minutes?)

src/apps/chifra/pkg/types/raw.go
```diff
@@ -0,0 +1,30 @@
+package types
```
tjayrush, line 1: I really like having a collection of "FromTheNode" and "FromTrueBlocks" data structures. Gives us a lot of flexibility. Maybe add a comment here?

src/apps/chifra/pkg/types/types.go
```diff
@@ -6,6 +6,20 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+type Modeler[Raw RawData] interface {
+	Model(showHidden bool, format string) Model
+	Raw() *Raw
+}
+
+type RawData interface {
+	RawReceipt
```
tjayrush, line 15: I'm curious -- will we be adding additional types here as we move forward? I think so, but not sure.

src/apps/chifra/pkg/types/types.go
```diff
@@ -105,32 +119,91 @@ type SimpleIndexAddressBelongs struct {
...
+func (r *SimpleReceipt) Raw() *RawReceipt {
+	return r.raw
+}
+
+func (r *SimpleReceipt) SetRaw(rawReceipt RawReceipt) {
+	r.raw = &rawReceipt
+}
+
+// To support custom format: just execute template on the output of Model
+func (r *SimpleReceipt) Model(showHidden bool, format string) Model {
```
tjayrush, line 160: General question (I'm not sure I know what I'm talking about, so don't change anything based on this question). Since we have the `raw` data, can we use pointers to the fields of that raw data in the Simple structures? Would that allow us to avoid copying bytes?

src/apps/chifra/pkg/utils/client.go
```diff
@@ -0,0 +1,7 @@
+package utils
+
+import "strings"
+
+func IsClientErigon(version string) bool {
```
tjayrush, line 8: Add a TODO item saying something like "Add queries for other node software (besu, nethermind, geth, etc." We will eventually need it anyway, I'm sure.

src/dev_tools/testRunner/testCases/tools/getReceipts.csv
```diff
@@ -33,9 +33,8 @@ on      ,both ,fast  ,receipts ,tools/getReceipts ,fmt_api             ,y    ,tr
 on      ,both ,fast  ,receipts ,tools/getReceipts ,fmt_json            ,y    ,transactions = 2222222.0 2332332.0 & fmt = json
 on      ,cmd  ,fast  ,receipts ,tools/getReceipts ,fmt_junk            ,y    ,transactions = 2222222.0 2332332.0 & fmt = junk
 on      ,cmd  ,fast  ,receipts ,tools/getReceipts ,by_filebad          ,n    ,file = cmd_file & bad
-on      ,cmd  ,fast  ,receipts ,tools/getReceipts ,by_file             ,n    ,file = cmd_file
-on      ,cmd  ,fast  ,receipts ,tools/getReceipts ,by_file_v           ,n    ,file = cmd_file & fmt = api & verbose
-on      ,both ,fast  ,receipts ,tools/getReceipts ,to_file             ,y    ,transactions = 2222222.0 2332332.0 & fmt = json & to_file
+off      ,cmd  ,fast  ,receipts ,tools/getReceipts ,by_file             ,n    ,file = cmd_file
```
tjayrush, line 36: Cool. We will come back to this (or any `off` test case).

src/apps/chifra/internal/receipts/output.go
```diff
@@ -45,12 +65,77 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
...
+		ShowKeys:   !opts.Globals.NoHeader,
+		ShowRaw:    opts.Globals.Raw,
+		ShowHidden: opts.Globals.Verbose,
+		Format:     opts.Globals.Format,
+		Meta:       meta,
+	})
+	if err != nil {
+		return err, true
+	}
+	return nil, true
```
tjayrush: Change this line to four lines thus:

```
    handled = true
    // EXISTING_CODE

    return
```

src/apps/chifra/pkg/types/types.go
dszlachta: Yes, I think we will have to. Although for raw types, we may decide to simply use `map[string]string`, so that we don't have to create new type for raw data. The disadvantage of it is that we wouldn't have type checking and non-existent keys would return `var, ok := raw["non-existent"] ---> var = ""; ok = false`. So if we use raw types in the code, we should have types for them, if not - then `map[string]string` should do the job.

src/apps/chifra/pkg/types/types.go
dszlachta: Yes, but: fields in SimpleTypes have different types than those in raw data. Everything in raw data is either a string or a slice of some type, whose fields are strings. This is how (Erigon at least) returns the data. So we need type casting to fill SimpleTypes with useful types (e.g. we probably want blockNumber to be a number instead of string)

src/apps/chifra/internal/receipts/output.go
dszlachta: Good point about using the package we already have 👍 , but `logger` also uses `os.Stderr`: https://github.com/TrueBlocks/trueblocks-core/blob/master/src/apps/chifra/pkg/logger/logger.go#L72
I think that our whole Go codebase expects an error to be returned and then it all works nice in both environments. Here we have a new situation, where there can be multiple errors. A quick and dirty workaround would be to just return the first error here. A better would be to try to change the console/server code to a generic version that expects `error` OR `[]error`

src/apps/chifra/internal/receipts/output.go
dszlachta: > Is there a reason we cannot use this routine: https://github.com/TrueBlocks/trueblocks-core/blob/master/src/apps/chifra/internal/receipts/options.go#L55?

That function reuses all of transaction ids and we need to send only some ids. That's why I created a new one. It's a workaround, but eventually we will remove it (when we port all of the code to Go)

src/apps/chifra/internal/receipts/output.go
dszlachta: I've put it above, if that's ok? I prefer to have constants defined at the top of a file

tjayrush: Unfortunately, this is an auto-generated file so the only place you can put non-boiler plate code is inside of `// EXISTING_CODE` blocks. Move this down to the bottom of the file and put a TODO item called // TODO: Add area for custom constants to top of file and I will fix that.

src/apps/chifra/internal/receipts/output.go
tjayrush: I see. Sounds good. You still have to move it so the auto-code generator doesn't blow it away. (We can get your setup to run the auto-code generator with each call to `make test`, and then you'll see how this works.

src/apps/chifra/internal/receipts/output.go
tjayrush: In the C++ code, the errors are reported to the server's console and in an object called `errors` (at the same level in the response as `data`)--so both stderr and in the response (i.e. stdout).

In command-line code, the error is reported only to stderr.

I'm going to resolve this conversation and write an issue.

tjayrush: https://github.com/TrueBlocks/trueblocks-core/issues/2339

src/apps/chifra/pkg/rpcClient/client.go
tjayrush: I wrote this issue: https://github.com/TrueBlocks/trueblocks-core/issues/2340

src/apps/chifra/pkg/types/types.go
tjayrush: I like types. I don't mind (in fact, I much prefer) to have types for nearly everything.

src/apps/chifra/pkg/types/types.go
tjayrush: I see. I thing our simple types (other than BigNumbers) should be native types. So when the JSON is produced for the front end, it's native. If you remember, BigInt (i.e. wei) is different. Closing for now.

Issue #2264, opened 2022-09-02
Detect when node is syncing


perama-v:
On TrueBlocks version 0.38.1
## Problem
When `chifra init --all` is run against a node that is syncing (such as Erigon Executing Blocks phase), the following error message appears:
```
Error: empty response  from provider http://localhost:8545 on chain mainnet
```
It is unclear if trueblocks is configure to talk to the node properly. Independently calling the node with `curl` may respond appropriately to some commands (net_version), implying (incorrectly) that there is a configuration problem on the trueblock side.
## Solution
If chifra receives a null response, it could query `eth_syncing`. If the response is not `false`, then `chifra init --all` could return
a message like: "Please wait for the node to finish syncing."

perama-v:
👏

PR #2265, opened 2022-09-02
Sync detect

perama-v:
Resolves user message issue [#2264](https://github.com/TrueBlocks/trueblocks-core/issues/2264) by
calling `eth_syncing` if `chifra init --all` returns a `nil` result.


tjayrush:
This is great. Thanks so much for this. I'm going to merge it, but as I do, I will add a TODO comment that this should really be done a bit higher level (for example inside the rpcClient code. This same error message could, conceivably apply to all the RPC calls. I'm not going to change it now because I want to make sure any change does not negatively affect the happy path, but the TODO note will allow me to find this issue later.

PR #2338, opened 2022-09-22
Feature/rpc counter

dszlachta:
This replaces manually set `ID` and `Jsonrpc` fields with auto-incremented, concurrency-safe counter and a static value

### DIFF
--- a/src/apps/chifra/pkg/rpcClient/rpc_cmds.go
+++ b/src/apps/chifra/pkg/rpcClient/rpc_cmds.go
@@ -9,11 +9,26 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
+	"sync/atomic"
 )
+
+var rpcCounter uint32
 
 // FromRpc Returns all traces for a given block.
 func FromRpc(rpcProvider string, payload *RPCPayload, ret interface{}) error {
-	plBytes, err := json.Marshal(payload)
+	type rpcPayload struct {
+		Jsonrpc   string `json:"jsonrpc"`
+		Method    string `json:"method"`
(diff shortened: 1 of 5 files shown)


src/apps/chifra/pkg/rpcClient/rpc_cmds.go
```diff
@@ -9,11 +9,26 @@ import (
...
+		Jsonrpc   string `json:"jsonrpc"`
+		Method    string `json:"method"`
+		RPCParams `json:"params"`
+		ID        int `json:"id"`
+	}
+	payloadToSend := rpcPayload{
+		Jsonrpc:   "2.0",
+		Method:    payload.Method,
+		RPCParams: payload.RPCParams,
+		ID:        int(atomic.AddUint32(&rpcCounter, 1)),
```
tjayrush, line 29: I'm pretty sure this serializes concurrent RPC requests.

If we want to do 10,000 RPC requests in parallel (not that we would, but if we wanted to, wouldn't each have to wait for the others to pass through this routine? I'm not sure that's worth it.

In seven years, I've never once used this -- even in very rare cases for debugging.

I would simply add one without forcing an atomic addition here.

dszlachta: If you "simply added one" here, you would get random results, because adding to ints is not safe with concurrency.
I did some extensive research and I've found that functions from `sync/atomic` package are translated directly into CPU instructions. I think this makes them very fast even for large number of operations. This package is also used internally by `sync` package, so if it was slow, it would make whole Go language slow.

PR #2372, opened 2022-09-29
Fix: set buffersize in chifra server Scan for progress

0xcybertim:
Fixes 'Error while reading stderr: bufio.Scanner: token too long' when exporting a block through the http api. For example block 4235012


tjayrush:
Thanks so much for your fix. I saw that error some time ago and couldn't figure it out. May I ask you to explain just a small bit what this does? Also -- why did you choose 1MB? Is that big enough? Too big? Best guess? (All three of these answers is fine, just curious.)

I'll merge this, but I'd appreciate your answers so I can semi-document this change.

PR #2381, opened 2022-09-30
getReceipts: Added support for output file

dszlachta:
I've added helper function that takes file path and a writer. If file path is not empty, the functions creates a file and returns new writer to this file. If the path is missing, it returns the writer that it gets as argument - that way we have simple code without too many `if`s.

This helper function is called directly from `StreamMany`, because it should decide how to print the output. In case we're running in server environment, `StreamMany` never calls the helper, so we always print to `http.ResponseWriter`. 

Right now file path is transported as a field in `OutputOptions`. You've mentioned that you're going to change `OutputOptions` and `GlobalOptions` and I didn't want to change code too deeply, introducing conflicts. This seemed like the most safe approach (and easy to change later)


src/apps/chifra/pkg/file/output.go
```diff
@@ -0,0 +1,25 @@
...
+// If there is an error, log.Fatal is called, because there really is no good way
+// to recover. Plus, output file is disabled in server, so it is safe to exit.
+func GetOutputFileWriter(filePath string, defaultWriter io.Writer) io.Writer {
+	if filePath == "" {
+		// If there's no file path, we can just return the default writer
+		return defaultWriter
+	}
+
+	// Otherwise, let's try to create the file
+	file, err := os.Create(filePath)
```
tjayrush, line 20: There's an Append option in the Globals data too. We use it in monitors. Should be a simple addition, just open the file with the right creation options.

Also, in general, I try to avoid using these 'convenience functions' from os package. If you look at this function (os.Create), it's a one-liner for ` OpenFile(name, O_RDWR|O_CREATE|O_TRUNC, 0666)`. In this way, I can search for O_RDWR and find all files opened for reading and writing, or O_CREATE vs O_APPEND to see which files we open for Append vs. destructive create.  I used this "ability to search" to make sure we are consistent in how we protect against destroying data (not finalized yet).

Upshot: use the expanded one-liner ` OpenFile(name, O_RDWR|O_CREATE|O_TRUNC, 0666)`

src/apps/chifra/pkg/file/output.go
```diff
@@ -0,0 +1,25 @@
...
+	if err != nil {
+		log.Fatal(err)
```
tjayrush, line 22: I'm curious about using Fatal -- does this kill the server if there's a problem? Not sure.

src/apps/chifra/pkg/utils/utils.go
```diff
@@ -27,6 +28,13 @@ func IsTestModeServer(r *http.Request) bool {
 	return r.Header.Get("User-Agent") == "testRunner"
 }
 
+// IsServerWriter tries to cast `w` into `http.ResponseWriter`
+// and returns true if the cast was successful
+func IsServerWriter(w io.Writer) bool {
```
tjayrush, line 33: I've been thinking that this exact function would be called `utils.IsApiMode()` which I've avoided up until now because I had to query the environment. I'd like to remove two things from the `Globals`: `ApiMode` and `fmt == "api"` if at all possible, but let's not worry about this just yet (I can do this with the change to the Global structure). Just make a TODO, I think for now.

src/apps/chifra/pkg/utils/utils.go
```diff
@@ -27,6 +28,13 @@ func IsTestModeServer(r *http.Request) bool {
...
+	_, ok := w.(http.ResponseWriter)
+	return ok
+}
+
 func IsApiMode() bool {
```
tjayrush, line 38: Oh. Just seeing this. I think the two (this and the above) are identical (although, I see that one needs the writer -- but I'm pretty sure we have that wherever we need it.

Upshot: don't do anything now other than note this.

src/apps/chifra/internal/receipts/output.go
```diff
@@ -104,13 +104,14 @@ func (opts *ReceiptsOptions) ReceiptsInternal() (err error, handled bool) {
 	}
 
 	err = output.StreamMany(ctx, opts.Globals.Writer, fetchTransactionData, output.OutputOptions{
```
tjayrush, line 106: The way the fix to Globals is working out, I think this will be the only line with a conflict. I'm going to merge this change first into `develop` and then merge into my code for the correction to Globals. Easier.

tjayrush:
Made a few comments. I pushed a change to remove the conflict from the change to Globals, so if you're ready to merge this, go ahead.

tjayrush:
I'm going to merge this as I'm getting a bit confused and I don't want to create too much of a mess. I'll resolve the above comments and make either issues or TODO items to keep track.

src/apps/chifra/pkg/utils/utils.go
tjayrush: I am going to try to remove utils.IsApiMode(). My guess is that it will be easy.

src/apps/chifra/pkg/file/output.go
tjayrush: I made this change.

src/apps/chifra/pkg/file/output.go
tjayrush: I added a // TODO:(dshlacta) comment

PR #2410, opened 2022-10-04
correct potenital typo

wschwab:
I think the changed line is supposed to read `init` instead of `index`, there is no `chifra index` on v40.0-beta, nor `chifra chunks index --all`, whereas there is `chifra init --all`, and later in the doc it seems to refer to this line as an initialization process.


tjayrush:
Yep. Thanks.

PR #2415, opened 2022-10-06
Feature/get receipts file argument

dszlachta:
Adds support for commands file


src/apps/chifra/cmd/receipts.go
```diff
@@ -24,7 +27,7 @@ var receiptsCmd = &cobra.Command{
 	Short:   shortReceipts,
 	Long:    longReceipts,
 	Version: versionText,
-	RunE:    receiptsPkg.RunReceipts,
+	RunE:    runWithFileSupport,
```
tjayrush: You may have though of this already, but perhaps `runWithFileSupport` could take a function pointer to `receiptsPkg.RunReceipts`. In that way, all commands could use the same `runWithFileSupport`.

src/apps/chifra/cmd/receipts.go
```diff
@@ -57,3 +60,46 @@ func init() {
...
+// but without wasting time and resources for the startup)
+func runWithFileSupport(cmd *cobra.Command, args []string) error {
+	// try to open the file
+	filePath, err := cmd.Flags().GetString("file")
+	if err != nil {
+		return err
+	}
+	if filePath == "" {
+		// `--file` has not been provided, run the command as usual
+		return receiptsPkg.RunReceipts(cmd, args)
```
tjayrush: This could be a function pointer so `runWithFileSupport` could be available to all commands.

src/apps/chifra/cmd/receipts.go
```diff
@@ -57,3 +60,46 @@ func init() {
...
+		}
+
+		// first, parse flags from the command line
+		_ = cmd.ParseFlags(os.Args[1:])
+		// next, parse flags from the file
+		err = cmd.ParseFlags(line.Flags)
+		if err != nil {
+			return fmt.Errorf("while parsing cmd flags: %s", err)
+		}
+		err = receiptsPkg.RunReceipts(cmd, line.Args)
```
tjayrush: This could be a function pointer so `runWithFileSupport` could be available to all commands.

src/apps/chifra/cmd/receipts.go
```diff
@@ -57,3 +60,46 @@ func init() {
...
+	}
+
+	// parse commands file
+	commandsFile, err := file.ParseCommandsFile(cmd, filePath)
+	if err != nil {
+		return err
+	}
+
+	for _, line := range commandsFile.Lines {
+		receiptsPkg.ResetOptions()
```
tjayrush: This could be a function pointer to make `runWtihFileSupport` available to all commands.

src/apps/chifra/pkg/file/commands.go
```diff
@@ -0,0 +1,70 @@
...
+	// TODO: parallelize
+	inputFile, err := os.Open(filePath)
+	if err != nil {
+		return
+	}
+
+	// scanner to get file lines
+	scanner := bufio.NewScanner(inputFile)
+	// we will use a simple counter to keep track of line numbers for improved error reporting
+	lineNumber := uint(0)
```
tjayrush, line 40: Nice idea. Helps the user with debugging.

tjayrush:
I made a few comments. Nothing earth shattering. Looks good.

src/apps/chifra/cmd/receipts.go
dszlachta: Yes, this is exactly what I want to do 👍

src/dev_tools/makeClass/templates/blank_options.go.tmpl
```diff
@@ -77,3 +77,10 @@ func GetOptions() *[{PROPER}]Options {
 	// EXISTING_CODE
 	return &default[{PROPER}]Options
 }
+
+func ResetOptions() {
```
tjayrush, line 81: Question for you: do you know how to run the auto code generator?

dszlachta: I think I do. Something along the lines of z`makeClass" "--options --gocmds`?

tjayrush: If it's configured, it should run every time you run either `make test-all` or `test-all.sh`. To configure, add this section to a file called `~/.local/share/trueblocks/makeClass.toml`.

```
[enabled]
generate = true
auto_format = true
readmes = true
```

tjayrush:
If you get a few minutes tomorrow morning run the 'auto code' generator. You'll see that four things happen:

1) All the other `options.go` files for each command will get the new `ResetOptions` code (you'll notice that your template has a hard-coded `Receipts` in it).
2) You'll see that the file.RunWithFileSupport line gets reset. This is because that code is in a file that is autogenerated, but you made a change outside of a `// EXISTING_CODE` section. All code outside of `// EXISTING_CODE` is overwritten when you run the code generator. You have to edit the corresponding template `./blank.go.tmpl`.
3) The third thing you'll find is that it won't compile because there's a missing import in each of the cmd files. You have to add the import lines for the `file` package by hand because that code is INSIDE the `// EXISTING_CODE` section.
4) Once you get it working, a bunch of the `--file` related test cases for other tools (`getBlocks`, etc.) will fail. It's perfectly okay (in fact, preferred) if you simply mark these newly failing cases as `local`. Fixing them is going to be hard, but only because the C++ code is messed up. We can fix it easier once this PR is merged.

To invoke the auto-code generator run `make generate finish`.

Note that if there's any code in any autogenerated file outside of `// EXISTING_CODE` sections, it will be lost if you don't have a backup -- it's unforgiving. I got a little bit burnt once or twice, so be careful.

If you don't feel like doing this, no problem, we can look at it tomorrow.

PR #2422, opened 2022-10-08
2418 add flow option to chifra transactions

tjayrush:


tjayrush:
Testing infrastructure is broken. Tests pass locally.

PR #2426, opened 2022-10-08
2425 fmt api is incorrect it should exist no where

tjayrush:


tjayrush:
This PR changes a large number of files include the new StreamMany routines, but it only does so to remove a huge confusion in the code. 

In summary, this PR removes the use of `api` as a `--fmt`. The only valid formats now are `txt`, `csv`, `json`, and a custom display string.

The confusion was caused by the fact that the existing code (including the new StreamMany code) was using `--fmt api` and `IsApiMode` interchangeably.

In previous changes, I consolidated `IsApiMode` to operate only by casting the writer on the `global` structure. This PR completes that work and separates `IsApiMode` and `--fmt` which really should have nothing to do with each other.

This modified the way the `meta` data is produced as well. Now, metaData is only produced in Api mode (and it's produced for all results from ApiMode) and it is not produced in non-Api mode.

Also, errors are included in any `json` mode whether Api or non-api.

PR #2437, opened 2022-10-13
Adds golang linting to ci

tjayrush:


tjayrush:
File added, but the action is turned off

tjayrush:
The file was added, but the action is turned off

PR #2473, opened 2022-10-21
Feature/2433 json writer

dszlachta:
`JsonWriter` is an abstraction that handles inserting opening and closing brackets, automatic comma insertion and indentation. Currently it is enabled only for `chifra receipts`, because only this code uses `StreamMany`. 
JsonWriter works both for CL (with or without `--file` and `--output`) and server.


tjayrush:
I'm going to wait to merge this until the test cases (such as ones labeled with `file_cmd` in this file: https://github.com/TrueBlocks/trueblocks-core/blob/feature/2433_json_writer/src/dev_tools/testRunner/testCases/tools/getReceipts.csv) labeled with `file_cmd` are working. All of the other test cases work, but I think they don't have any examples of multiple JSON objects with the same format.

src/apps/chifra/internal/abis/output.go
```diff
@@ -20,18 +21,23 @@ import (
 // RunAbis handles the abis command for the command line. Returns error only as per cobra.
 func RunAbis(cmd *cobra.Command, args []string) (err error) {
 	opts := abisFinishParse(args)
+	outputHelpers.InitJsonWriter("abis", &opts.Globals)
```
tjayrush: Can we move this line....

src/apps/chifra/internal/abis/output.go
```diff
@@ -20,18 +21,23 @@ import (
...
 	// EXISTING_CODE
 	// EXISTING_CODE
 	err, _ = opts.AbisInternal()
+	outputHelpers.CloseJsonWriterIfNeeded("abis", &opts.Globals)
```
tjayrush: ....and this line....

src/apps/chifra/internal/abis/output.go
```diff
@@ -20,18 +21,23 @@ import (
...
 	return
 }
 
 // ServeAbis handles the abis command for the API. Returns error and a bool if handled
 func ServeAbis(w http.ResponseWriter, r *http.Request) (err error, handled bool) {
 	opts := abisFinishParseApi(w, r)
+	outputHelpers.InitJsonWriterApi("abis", w, &opts.Globals)
```
tjayrush, line 34: ....and this line....

src/apps/chifra/internal/abis/output.go
```diff
@@ -20,18 +21,23 @@ import (
...
 	// EXISTING_CODE
 	// EXISTING_CODE
-	return opts.AbisInternal()
+	err, handled = opts.AbisInternal()
```
tjayrush, line 37: ...and these lines....

src/apps/chifra/internal/abis/output.go
```diff
@@ -20,18 +21,23 @@ import (
...
+	outputHelpers.CloseJsonWriterIfNeededApi("abis", err, &opts.Globals)
+	return
```
tjayrush, line 39: down into AbisInternal (and the associated routines under each command) and combine the four functions into two that use the fact that we know ApiMode()?

It lessens the cognitive load because, if I'm not mistaken, these routines do the same thing just slightly different.

Perhaps you didn't do this for a reason? Maybe the return values on error?

Seems cleaner to me.

Maybe something like `InitJsonWriter` whose first line is:

```
if globals.IsApiMode() {
    return InitJsonWriterApi(....)
}
```

If there's a reason why it's like this, I'm open to discussion.

src/apps/chifra/internal/explore/output.go
```diff
@@ -25,18 +26,23 @@ import (
 // RunExplore handles the explore command for the command line. Returns error only as per cobra.
 func RunExplore(cmd *cobra.Command, args []string) (err error) {
 	opts := exploreFinishParse(args)
+	outputHelpers.InitJsonWriter("explore", &opts.Globals)
 	// EXISTING_CODE
 	// EXISTING_CODE
 	err, _ = opts.ExploreInternal()
```
tjayrush, line 32: Can you please add a TODO: item here saying "TODO: does chifra explore need to use the StreamMany stuff? This is a sort of special case, but we can test it later.

src/apps/chifra/internal/globals/options.go
```diff
@@ -27,6 +27,10 @@ type GlobalOptions struct {
 	Version bool   `json:"version,omitempty"`
 	Noop    bool   `json:"noop,omitempty"`
 	NoColor bool   `json:"noColor,omitempty"`
+	// This flag is used in combination with --file. It informs,
+	// that we created JsonWriter that should not be passed to
+	// the next command from the file.
+	MixedFormats bool `json:"mixedFormats,omitempty"`
```
tjayrush: There's an issue in our docs repo saying that we explicitly refuse to support this (mixed formats). I wonder if the complication is (if there actually is any) is worth it. Would people actually do this (mix formats)?

Although, now that I think of it, perhaps someone would do something likes this:

```
chifra export <some options> --fmt json --output file.json
chifra export <some options> --fmt csv --output file.csv
```

In other words, produce data in two formats for the exact same options. So, I guess, keep it.

src/apps/chifra/pkg/output/helpers/helpers.go
```diff
@@ -0,0 +1,94 @@
...
+	"os"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/internal/globals"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/output"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/rpcClient"
+	"github.com/spf13/cobra"
+)
+
+var enabledForCmds = map[string]bool{
+	"receipts": true,
```
tjayrush, line 14: Add a TODO item here called "This list can eventually be auto generated."

src/apps/chifra/internal/abis/options.go
```diff
@@ -136,6 +137,16 @@ func GetOptions() *AbisOptions {
 }
 
 func ResetOptions() {
+	// We want to keep writer between command file calls if
+	// the global format is JSON
+	w := GetOptions().Globals.Writer
 	defaultAbisOptions = AbisOptions{}
 	globals.SetDefaults(&defaultAbisOptions.Globals)
+	if GetOptions().Globals.Format == "json" || GetOptions().Globals.ShowRaw {
```
tjayrush: Correct me if I'm wrong, but I think this test appears in numerous other places, and its purpose is basically to say "the following section of code is 'different' in the sense that it needs to 'close'" with "close" meaning that it must write the closing bracket for JSON, but in the case of a custom format could mean "put a date on the bottom of the custom built string that is the bank statement."

Can we make this test (wherever it appears) be a routine called something like `needsFooter` (that why I used `RenderHeader` and `RenderFooter` in the other code.

src/apps/chifra/pkg/output/helpers/helpers.go
```diff
@@ -0,0 +1,94 @@
...
+		jw.ShouldWriteMeta = true
+		jw.GetMeta = func() (*rpcClient.MetaData, error) {
+			return rpcClient.GetMetaData(opts.Chain, opts.OutputOptions.TestMode)
+		}
+		opts.Writer = jw
+	}
+}
+
+func InitJsonWriter(cmdName string, opts *globals.GlobalOptions) {
+	if !enabledForCmds[cmdName] {
```
tjayrush: Couldn't the first line of this be:

```
if opts.IsApiMode() {
    InitJsonWriterApi(cmdName, opts)
    return
}
```

src/apps/chifra/pkg/output/helpers/helpers.go
```diff
@@ -0,0 +1,94 @@
...
+		// to the next command (next line of --file)
+		opts.MixedFormats = true
+	}
+	if !ok && !(opts.Format == "json" || opts.ShowRaw) {
+		opts.Writer = opts.GetOutputFileWriter()
+	}
+}
+
+func CloseJsonWriterIfNeeded(cmdName string, opts *globals.GlobalOptions) {
+	if enabledForCmds[cmdName] && (opts.Format == "json" || opts.ShowRaw) && opts.MixedFormats {
```
tjayrush: Couldn't the first line of this be:

```
if opts.IsApiMode() {
    CloseJsonWriterIfNeededApi(cmdName, opts)
    return
}
```

src/apps/chifra/pkg/output/json_writer.go
```diff
@@ -0,0 +1,332 @@
...
+	indentLevel  int
+	indentString string
+	// errors to output
+	errs []string
+	// function to get meta data
+	GetMeta func() (*rpcClient.MetaData, error)
+	// flag indicating if we should output `meta` object as
+	// well
+	ShouldWriteMeta bool
+	// Should writer write newline after `Close`
```
tjayrush, line 61: Maybe this is always `true`? Is the only reason for this in order to agree with test cases, or is there some other reason?

src/apps/chifra/pkg/output/json_writer.go
```diff
@@ -0,0 +1,332 @@
...
+	bw, err := w.writeNewline()
+	n += bw
+	if err != nil {
+		return
+	}
+	w.Indent()
+	if w.state.position == positionInArray || w.state.position == positionInObject || w.state.position == positionInRoot {
+		w.state.children++
+	}
+	if err != nil {
```
tjayrush: I think we've already tested this...

src/apps/chifra/pkg/output/json_writer.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	return w.Write([]byte(fmt.Sprintf(`"%s": %s`, key, value)))
+}
+
+// WriteError saves error to be written when the writer is `Close`d
+func (w *JsonWriter) WriteError(err error) {
+	w.errs = append(w.errs, err.Error())
+}
+
+// WriteCompoundItem makes it easier to writer an object or array.
```
tjayrush: "easier to >>write<< an object or array

src/apps/chifra/pkg/output/stream.go
```diff
@@ -57,11 +55,15 @@ func StreamWithTemplate(w io.Writer, model types.Model, tmpl *template.Template)
 // StreamModel streams a single `Model`
 func StreamModel(w io.Writer, model types.Model, options OutputOptions) error {
 	if options.Format == "json" {
```
tjayrush, line 57: Why, in some cases, do we test only for `Format == "json"` and in other we test both `Format == "json" || opts.IsRaw`? I would prefer if we forced `Format = "json"` if we parsed `--raw` from the command line. Is that possible?

src/apps/chifra/pkg/output/stream.go
```diff
@@ -187,18 +163,18 @@ func StreamMany[Raw types.RawData](ctx context.Context, fetchData fetchDataFunc[
 
 			// If the output is JSON and we are printing another item, put `,` in front of it
 			// TODO: BOGUS - ShowRaw is json
-			if !first && isJson {
-				outputWriter.Write([]byte(","))
-			}
+			// if !first && isJson {
```
tjayrush: Remove comments

src/apps/chifra/pkg/output/stream_test.go
```diff
@@ -87,13 +88,20 @@ func TestStreamFormats(t *testing.T) {
 
 func TestStreamJson(t *testing.T) {
 	outputBuffer := &bytes.Buffer{}
-	StreamModel(outputBuffer, input.Model(false, "json"), OutputOptions{
+	w := NewJsonWriter(outputBuffer)
+	w.DefaultField = DefaultField{Key: "Data", FieldType: FieldArray}
```
tjayrush, line 92: I'm not sure, but should "Data" be lower case?

src/apps/chifra/pkg/output/json_writer.go
dszlachta: Good catch

src/apps/chifra/pkg/output/stream.go
dszlachta: We need to check both conditions if the code is running before `InitGlobals` (which sets format to `"json"` always when `--raw` is present)

PR #2481, opened 2022-10-23
2474 - Fix scrape crash

marsu-dev:
The fix ensure slice pre-allocation with original BlockCnt

Rework
- The `GOTO` statement is removed, and a chan timer is used to wait for next iteration.
- Inner loop for pause computation is removed
- Loggin is  enhanced, some errors was not logged.
- the `HandleScrape` still can return unexpectedly

### DIFF
--- a/src/apps/chifra/internal/scrape/handle_scrape.go
+++ b/src/apps/chifra/internal/scrape/handle_scrape.go
@@ -78,8 +78,8 @@ func (opts *ScrapeOptions) HandleScrape() error {
 			UnripeDist:    opts.Settings.Unripe_dist,
 			RpcProvider:   config.GetRpcProvider(opts.Globals.Chain),
 			AppearanceMap: make(index.AddressAppearanceMap, opts.Settings.Apps_per_chunk),
-			TsArray:       make([]tslib.Timestamp, 0, opts.BlockCnt),
-			ProcessedMap:  make(map[int]bool, opts.BlockCnt),
+			TsArray:       make([]tslib.Timestamp, 0, origBlockCnt),
+			ProcessedMap:  make(map[int]bool, origBlockCnt),
 		}
 
 		// Remove whatever's in the unripePath before running each round. We do this
--- a/src/apps/chifra/internal/scrape/handle_scrape.go
+++ b/src/apps/chifra/internal/scrape/handle_scrape.go
@@ -5,6 +5,7 @@ package scrapePkg
 // be found in the LICENSE file.
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -24,6 +25,7 @@ import (
 func (opts *ScrapeOptions) HandleScrape() error {
 	progress, err := rpcClient.GetMetaData(opts.Globals.Chain, opts.Globals.TestMode)
 	if err != nil {
+		logger.Log(logger.Error, err)
 		return err
 	}
 
@@ -41,6 +43,7 @@ func (opts *ScrapeOptions) HandleScrape() error {
 	}
 
 	if ok, err := opts.HandlePrepare(progress, &blazeOpts); !ok || err != nil {
+		logger.Log(logger.Error, err)
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


src/apps/chifra/internal/scrape/handle_scrape.go
```diff
@@ -115,12 +122,10 @@ func (opts *ScrapeOptions) HandleScrape() error {
 			if !ok {
 				break
 			}
-			goto PAUSE
```
tjayrush, line 118: First. Thanks so much for your PR. The intention of this is that the scraper never quits. It pauses under the assumption that the node software's RPC has temporarily had a problem. In other words, even if the node software dies, I think the scraper should not exit it's loop.

src/apps/chifra/internal/scrape/handle_scrape.go
```diff
@@ -115,12 +122,10 @@ func (opts *ScrapeOptions) HandleScrape() error {
...
+			continue
 		}
-
-	PAUSE:
-		opts.Pause(progress)
 	}
 
-	return nil
+	// should never append (see caller)
+	return errors.New("unexpected loop exit")
```
tjayrush, line 130: There are many cases where the loop will return an error (for example, the node software died). We shouldn't leave the loop simply because of an error. Perhaps there's a better way to distinguish why the error occurred, but simply leaving this loop isn't the right solution, I don't think.

src/apps/chifra/internal/scrape/pause.go
```diff
@@ -7,21 +7,32 @@ import (
...
 )
 
-func (opts *ScrapeOptions) Pause(progressThen *rpcClient.MetaData) {
-	// we always pause at least a quarter of a second to allow the node to 'rest'
-	time.Sleep(250 * time.Millisecond)
+const (
+	minimumSleep = time.Second / 4
+)
+
+func (opts *ScrapeOptions) Pause(progressThen *rpcClient.MetaData) <-chan time.Time {
```
tjayrush, line 14: Thanks for fixing this. This Pause code has always been a bit of troubling. Can you add a comment here explaining (briefly) its operation?

src/apps/chifra/internal/scrape/pause.go
```diff
@@ -7,21 +7,32 @@ import (
...
+	sleep := opts.computePause(progressThen)
+	return time.After(sleep)
```
tjayrush, line 16: I wasn't aware of this routine. Very clear. Thanks.

src/apps/chifra/internal/scrape/pause.go
```diff
@@ -7,21 +7,32 @@ import (
...
+
 	isDefaultSleep := opts.Sleep >= 13 && opts.Sleep <= 14
 	distanceFromHead := progressThen.Latest - progressThen.Staging
 	shouldSleep := !isDefaultSleep || distanceFromHead <= (2*opts.Settings.Unripe_dist)
 	if shouldSleep {
-		sleep := opts.Sleep
-		if sleep > 1 {
-			logger.Log(logger.Info, "Sleeping for", sleep, "seconds -", distanceFromHead, "away from head.")
-		}
-		halfSecs := (sleep * 2) - 1 // we already slept one quarter of a second
```
tjayrush, line 21: This was here for a reason. In a process completely outside of this sleep loop, it's possible for some external code to signal that the scraper should (a) restart, (b) quit, (c) pause. If the user, for example, makes `Sleep` = 1000, then the scraper won't respond to those signals for 1000 seconds. If we come back alive every half second, the response is "snappy".

tjayrush (changes requested):

tjayrush:
Thanks for your PR. We appreciate it. I made some comments.

tjayrush:
I am going to close the PR due to a non-response to my comments. If you wish, you may re-open it, but if you do, please address my comments. 

Thanks so much for bringing this to our attention. We won't forget it, but it will get incorporated once we have a chance to re-visit the scraper code.

PR #2584, opened 2022-11-24
Cache support

dszlachta:
This PR implements caching. All of the code is in `pkg/cache`. When reading and writing, we use `Simple*` types.


src/apps/chifra/pkg/cache/paths.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+	"path"
+	"strings"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
+)
+
+var itemToDirectory = map[CacheItem]string{
```
tjayrush, line 11: I wrote code similar to this recently, but I like your code better. I anticipated this, so I have no problem removing my code and using this. We can do that after we merge. There's a slightly different way of figuring out `path` for `monitors` and `reconciliations`, but they will easily fit into this structure. Looks good.

src/apps/chifra/pkg/cache/cache.go
```diff
@@ -0,0 +1,135 @@
...
+	bufReader := bufio.NewReader(reader)
+	value, err = read(bufReader)
+	if err != nil && !os.IsNotExist(err) {
+		// Ignore the error, we will re-try next time
+		remove(chainName, filePath)
+	}
+	return
+}
+
+// SetBlock stores block in the cache
```
tjayrush, line 91: I think in the future we will want to move the 'type' specific code into a type specific file for block, transaction, etc. In this way, this mostly boilerplate code can be auto-generated.

Don't change anything other than perhaps to add a TODO item here saying something about "Move to it's own type specific file for the block structure."

Applies to the Transaction code below and the data specific code in `cache/read.go` (I put a comment there too.)

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+
+	err = readCString(reader, &target.className)
+	if err != nil {
+		return err
+	}
+
+	return
+}
+
+func ReadBlock(reader *bufio.Reader) (block *types.SimpleBlock, err error) {
```
tjayrush, line 224: I feel like eventually we will want to move this data type specific code to a data type specific file. In this way, it can be more easily auto-generated. Don't change this, but please add a TODO list item saying something similar to the note I asked you to add above (So I can search for and find both with one search.)

Same for ReadTransaction below.

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+	articulatedTx, err := ReadFunction(reader)
+	if err != nil {
+		return
+	}
+	tx.ArticulatedTx = articulatedTx
+
+	return
+}
+
+func ReadReceipt(reader *bufio.Reader) (receipt *types.SimpleReceipt, err error) {
```
tjayrush, line 430: Eventually, we'll move this into the auto-generated type specific file. Please add a TODO, but don't move this yet.

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+	}
+
+	err = read(&receipt.Status)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
```
tjayrush, line 471: Eventually, we'll move this into the auto-generated type specific file. Please add a TODO, but don't move this yet.

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+	err = readFromArray(reader, &block.Transactions, ReadTransaction)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
+func ReadTransaction(reader *bufio.Reader) (tx *types.SimpleTransaction, err error) {
```
tjayrush, line 296: Eventually, we'll move this into the auto-generated type specific file. Please add a TODO, but don't move this yet.

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+	articulatedLog, err := ReadFunction(reader)
+	if err != nil {
+		return
+	}
+	log.ArticulatedLog = articulatedLog
+
+	return
+}
+
+func ReadFunction(reader *bufio.Reader) (function *types.SimpleFunction, err error) {
```
tjayrush, line 514: Eventually, we'll move this into the auto-generated type specific file. Please add a TODO, but don't move this yet.

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+
+	err = readFromArray(reader, &function.Outputs, ReadParameter)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
+func ReadParameter(reader *bufio.Reader) (param *types.SimpleParameter, err error) {
```
tjayrush, line 580: Eventually, we'll move this into the auto-generated type specific file. Please add a TODO, but don't move this yet.

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+
+	err = read(&param.IsFlags)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
+func ReadTrace(reader *bufio.Reader) (trace *types.SimpleTrace, err error) {
```
tjayrush, line 641: Eventually, we'll move this into the auto-generated type specific file. Please add a TODO, but don't move this yet.

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+	function, err := ReadFunction(reader)
+	if err != nil {
+		return
+	}
+	trace.ArticulatedTrace = function
+
+	return
+}
+
+func readTraceAction(reader *bufio.Reader) (action *types.SimpleTraceAction, err error) {
```
tjayrush, line 715: Eventually, we'll move this into the auto-generated type specific file. Please add a TODO, but don't move this yet.

src/apps/chifra/pkg/cache/read.go
```diff
@@ -0,0 +1,813 @@
...
+
+	err = readBigUint(reader, &action.Value)
+	if err != nil {
+		return
+	}
+
+	return
+}
+
+func readTraceResult(reader *bufio.Reader) (result *types.SimpleTraceResult, err error) {
```
tjayrush, line 781: Eventually, we'll move this into the auto-generated type specific file. Please add a TODO, but don't move this yet.

src/apps/chifra/pkg/cache/types.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+type cacheHeader struct {
+	deleted   uint64
+	schema    uint64
+	showing   uint64
+	className cString
+}
+
+func reverseBytes(original []byte) (reversed []byte) {
```
tjayrush, line 15: Can you explain this? Is it needed? What for?

src/apps/chifra/pkg/cache/types.go
```diff
@@ -0,0 +1,25 @@
+package cache
+
+type unsignedLong = uint64
```
tjayrush: Do we need this?

src/apps/chifra/pkg/cache/types.go
dszlachta: Sure, I was surprised that we need it, too. So, when Go sees `[]byte`, it ignores byte order when using `binary.Read` and `binary.Write`. It treats `[]byte` as raw bytes. And because our files are little endian and Go is big endian, I have to reverse this slice

src/apps/chifra/pkg/cache/types.go
dszlachta: It's not vital in any way. It just makes `type cString` definition look like the one in C++.

src/apps/chifra/pkg/cache/types.go
tjayrush: There are GoLang routines to specify the endian (for example, here: https://github.com/TrueBlocks/trueblocks-core/blob/master/src/apps/chifra/pkg/index/chunk_data_appearance_record.go#L48.)

Could we not have used that?

I'd prefer to be very specific so there's no ambiguity.

Is that possible?

src/apps/chifra/pkg/cache/types.go
tjayrush: Let's remove it.

src/apps/chifra/pkg/cache/types.go
dszlachta: We are using them (check `createReadFn`). Here's the issue: https://github.com/golang/go/issues/40891

PR #2599, opened 2022-12-02
Added file lock for cache

dszlachta:
This is OS lock, which means it works across processes. Inside of the same process, but in different goroutines, we're covered by Go's thread-safe `os.File`.

### DIFF
--- a/src/apps/chifra/pkg/cache/cache.go
+++ b/src/apps/chifra/pkg/cache/cache.go
@@ -31,9 +31,9 @@ func save(chain string, filePath string, content io.Reader) (err error) {
 	fullPath := path.Join(cacheDir, filePath)
 
 	var file *os.File
-	if _, err = os.Stat(fullPath); os.IsNotExist(err) {
+	if filePkg.FileExists(fullPath) {
 		// If file doesn't exist, we don't need a lock
-		file, err = os.Create(fullPath)
+		file, err = os.OpenFile(fullPath, os.O_RDWR|os.O_CREATE, 0666)
 		if err != nil {
 			return
 		}
--- a/src/apps/chifra/pkg/cache/cache.go
+++ b/src/apps/chifra/pkg/cache/cache.go
@@ -8,6 +8,7 @@ import (
 	"path"
 
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
+	filePkg "github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/file"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
 )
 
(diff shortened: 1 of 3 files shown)


src/apps/chifra/pkg/file/empty.go
```diff
@@ -0,0 +1,20 @@
...
+// be found in the LICENSE file.
+
+package file
+
+import "os"
+
+// Empty is a shortcut function to seek to the beginning
+// of a file and truncate it to 0. It does not call
+// file.Sync though, making all operations in-memory only.
+func Empty(file *os.File) (err error) {
```
tjayrush, line 12: Can't you just open the file with the right mode? os.O_RDWR|os.O_CREATE|os.O_TRUNC

src/apps/chifra/pkg/cache/cache.go
```diff
@@ -28,17 +29,32 @@ func getCacheAndChainPath(chain string) string {
 func save(chain string, filePath string, content io.Reader) (err error) {
 	cacheDir := getCacheAndChainPath(chain)
 	fullPath := path.Join(cacheDir, filePath)
-	file, err := os.Create(fullPath)
-	if err != nil {
-		return
+
+	var file *os.File
+	if _, err = os.Stat(fullPath); os.IsNotExist(err) {
```
tjayrush: There's a utility in our `file` package called `file.FileExists()` that returns a boolean.

src/apps/chifra/pkg/cache/cache.go
```diff
@@ -28,17 +29,32 @@ func getCacheAndChainPath(chain string) string {
...
+		// If file doesn't exist, we don't need a lock
+		file, err = os.Create(fullPath)
```
tjayrush: I'm not in love with using GoLang's shorthand functions for opening files. If you look at that code, it just does:

```
OpenFile(name, O_RDWR|O_CREATE|O_TRUNC, 0666)
```

I'd rather have this explicit code in our code so I can see exactly what's happening without having to know how the GoLang code works. In fact, about three months ago, I removed all the "shorthand" code so I can search for `OpenFilename` and see everywhere I'm opening a file of any type.

src/apps/chifra/pkg/cache/cache.go
```diff
@@ -28,17 +29,32 @@ func getCacheAndChainPath(chain string) string {
...
+		if err != nil {
+			return
+		}
+		defer file.Close()
```
tjayrush, line 40: This is just a question for my own education. I C++ this line would execute at the end of this code block (if c++ had `defer`), but in GoLang, `defer` is executed at the end of the function call. Is that correct? Is that correct in every case?

src/apps/chifra/pkg/file/lock.go
```diff
@@ -49,3 +50,28 @@ func changeLock(file *os.File, lockConfig *syscall.Flock_t) error {
...
+			// Another process has already locked the file, we wait
+			continue
+		}
+		if err != nil {
+			// We cannot lock, so report the error back to the caller
+			return file, err
+		}
+		time.Sleep(1 * time.Second)
+	}
+	return
```
tjayrush: If I'm reading this right, we've failed for three seconds to get the lock. Shouldn't this be an error return?

src/apps/chifra/pkg/file/empty.go
dszlachta: The thing is that this would truncate the file when opening it. And I need to check for the lock first. Locks are additional protocol and `file.Open` doesn't care about them.

src/apps/chifra/pkg/cache/cache.go
dszlachta: Noted 👍

src/apps/chifra/pkg/cache/cache.go
dszlachta: Yes, this is correct for every case.

src/apps/chifra/pkg/file/lock.go
dszlachta: Thank you for spotting this. Good catch

tjayrush:
Looks good. I guess I'll merge it once the tests work.

tjayrush:
Fixed the testing. There was something wrong with Python on the testing machine. I change the CI to use python3 directly.

tjayrush:
Let me know if you want to merge this. It says it's a WIP.

dszlachta:
I will just merge it

PR #2626, opened 2022-12-15
Feature/docker version

tjayrush:


tjayrush:
Rewrote as new-docker-version

PR #2655, opened 2023-01-03
More flexible Model definition

dszlachta:
In order to support two different types of `SimpleBlock.Transactions` (`[]string` or `[]SimpleTransaction`) I had to add generic parameter SimpleBlock. 

This forces us to have two different function for getting a block from RPC, so that Go knows what concrete type it gets. But I think it's OK.

SimpleTransaction is now `Modeler`, so it has `Model` method. This method is called internally by Block's Model, if transactions are objects (not strings).

Fixes for some tests (mainly displaying data that was previously missing) are also in this PR.

~~`Model`'s `Data()` method will now return value of `any` type. As you know, I'm not a big fan of `any`, but in this particular case all we do with the returned value is pass it to `JsonWriter` or CSV/TSV encoders or `fmt.Sprint` which all accept `any` value. So, there's no drawbacks here.~~

~~`chifra blocks --hashes` (and probably other commands as well) forces us to have a way to pass command-specific flags to the model (e.g. *hey, only print hashes now*). To do so I had to add another parameter to `Model.Data()` method, named `extraOptions` and add a new field to `OutputOptions`. I think that we both agree that `OutputOptions` can be done in a better way, but that's a refactoring we should plan and I didn't want to commit big changes to this branch.~~


src/apps/chifra/internal/blocks/handle_show.go
```diff
@@ -61,5 +61,8 @@ func (opts *BlocksOptions) HandleShowBlocks() error {
 		OutputFn:   opts.Globals.OutputFn,
 		Append:     opts.Globals.Append,
 		JsonIndent: "  ",
+		ModelExtraOptions: map[string]interface{}{
```
tjayrush: Can we just call this 'Extra'? We're going to rewrite the way this works, but it just looks out of place to me. Sorry to be anal.

tjayrush:
This works when I build it and run it on the command line, but the tests fail. Can you fix the test and re-publish? I'll merge it.

src/apps/chifra/pkg/types/types_test.go
```diff
@@ -12,19 +12,21 @@ func TestTypes(t *testing.T) {
...
 		t.Fatal("hidden field leaked")
 	}
 
-	if r.Model(false, "csv").Data["blockNumber"] == nil {
+	if data["blockNumber"] == nil {
 		t.Fatal("visible field missing")
 	}
 
-	if r.Model(true, "json").Data["blockHash"] == nil {
+	if data["blockHash"] == nil {
```
tjayrush, line 29: This test fails. Can we use this as a learning opportunity? Is this related to the conversion to `any` type?

tjayrush:
Also, you may have to re-base (although I'm not even sure what that means).

src/apps/chifra/pkg/types/types_test.go
dszlachta: No, it was my mistake. This part of the test was calling `r.Model(true, "json")` before to force showing the hidden fields. I forgot to change these two arguments

dszlachta:
> Also, you may have to re-base (although I'm not even sure what that means).

Done.

tjayrush:
Sorry. Do this:

```
TEST_TEST_ONLY=false chifra blocks 1001010 | jq -S >file.1
TEST_TEST_ONLY=true chifra blocks 1001010 | jq -S >file.2
diff file.1 file.2
```

You'll see that the new code shows `nil` for the transactions field.

Also,

```
TEST_TEST_ONLY=false chifra blocks 1001010 --hashes | jq -S >file.1
TEST_TEST_ONLY=true chifra blocks 1001010 --hashes | jq -S >file.2
diff file.1 file.2
```

The new code only shows hashes but not the rest of the fields in the block.

PR #2661, opened 2023-01-09
Feature/port chifra abis

dszlachta:
This is work in progress. Do not merge.


src/apps/chifra/pkg/abi/load_abi.go
```diff
@@ -0,0 +1,236 @@
...
+var knownAbiSubdirectories = []string{
+	"known-000", "known-005", "known-010", "known-015",
+}
+
+type AbiInterfaceMap = map[string]*types.SimpleFunction
+
+// getMapKey turns encoding into lower case, which can be used
+// as a key in AbiInterfaceMap
+func getMapKey(encoding string) string {
+	return strings.ToLower(encoding)
```
tjayrush, line 31: Can we consolidate this in the same way as the Address thing? Seems similar.

src/apps/chifra/pkg/abi/load_abi.go
```diff
@@ -0,0 +1,236 @@
...
+func fromJson(reader io.Reader, abiSource string, destination AbiInterfaceMap) (err error) {
+	// Compute encodings, signatures and parse file
+	loadedAbi, err := abi.JSON(reader)
+	if err != nil {
+		return
+	}
+
+	for _, method := range loadedAbi.Methods {
+		function := types.FunctionFromAbiMethod(&method, abiSource)
+		destination[getMapKey(function.Encoding)] = function
```
tjayrush, line 54: Is it possible to make sure the Encoding is already lower case when it's loaded? Not sure that makes sense, just curious.

src/apps/chifra/pkg/abi/load_abi.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	for _, ethEvent := range loadedAbi.Events {
+		event := types.FunctionFromAbiEvent(&ethEvent, abiSource)
+		destination[getMapKey(event.Encoding)] = event
```
tjayrush, line 59: Is it possible to make sure the Encoding is already lower case when it's loaded? Not sure that makes sense, just curious.

src/apps/chifra/pkg/abi/load_abi.go
```diff
@@ -0,0 +1,236 @@
...
+
+	var functions []types.SimpleFunction
+	decoder := json.NewDecoder(file)
+	err = decoder.Decode(&functions)
+	if err != nil {
+		return
+	}
+
+	for _, function := range functions {
+		destination[getMapKey(function.Encoding)] = &function
```
tjayrush, line 81: Is it possible to make sure the Encoding is already lower case when it's loaded? Not sure that makes sense, just curious.

src/apps/chifra/pkg/abi/load_abi.go
```diff
@@ -0,0 +1,236 @@
...
+func LoadCache(chain string, destination AbiInterfaceMap) (loaded bool) {
+	functions, cacheErr := cache.GetAbis(chain)
+	// We can ignore cache error
+	if cacheErr != nil {
+		return
+	}
+
+	for _, function := range functions {
+		function := function
+		destination[getMapKey(function.Encoding)] = &function
```
tjayrush, line 107: Is it possible to make sure the Encoding is already lower case when it's loaded? Not sure that makes sense, just curious.

src/apps/chifra/pkg/abi/load_abi.go
```diff
@@ -0,0 +1,236 @@
...
+
+		filePaths = append(filePaths, path)
+		return nil
+	})
+	return
+}
+
+// LoadAbiFromAddress loads ABI from local file or cache
+func LoadAbiFromAddress(chain string, address common.Address, destination AbiInterfaceMap) (err error) {
+	localFileName := strings.ToLower(address.Hex()) + ".json"
```
tjayrush, line 206: The Address thing

src/apps/chifra/pkg/abi/load_abi.go
```diff
@@ -0,0 +1,236 @@
...
+	localFile, err := os.OpenFile(localFileName, os.O_RDONLY, 0)
+	if os.IsNotExist(err) {
+		// There's no local file, so we try to load one from cache
+		loadedAbis, err := cache.GetAbi(chain, address)
+		if err != nil {
+			return err
+		}
+		for _, loadedAbi := range loadedAbis {
+			loadedAbi := loadedAbi
+			destination[getMapKey(loadedAbi.Encoding)] = &loadedAbi
```
tjayrush, line 216: Is it possible to make sure the Encoding is already lower case when it's loaded? Not sure that makes sense, just curious.

src/apps/chifra/pkg/abi/load_abi.go
```diff
@@ -0,0 +1,236 @@
...
+	toCache := make([]types.SimpleFunction, 0, len(destination))
+	for _, function := range destination {
+		function := function
+		toCache = append(toCache, *function)
+	}
+	return cache.SetAbis(chain, toCache)
+}
+
+// findKnownAbi finds known ABI by name in known-* directories
+func findKnownAbi(name string) (filePath string) {
```
tjayrush, line 166: Question: where is this function called from?

src/apps/chifra/internal/abis/handle_addresses.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/output"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+func (opts *AbisOptions) HandleAddresses() (err error) {
+	testMode := opts.Globals.TestMode
+	result := make(abi.AbiInterfaceMap)
+	if opts.Known {
+		if err = abi.PreloadKnownAbis(opts.Globals.Chain, result, false); err != nil {
```
tjayrush, line 20: I'm confused by this a little bit. The "Handle" routines typically generate output, but doesn't this return without having generated output? If yes, my question is why do we come in here at all? Why not do this test before calling?

src/apps/chifra/internal/abis/handle_addresses.go
```diff
@@ -0,0 +1,113 @@
...
+			return
+		}
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	fetchData := func(modelChan chan types.Modeler[types.RawFunction], errorChan chan error) {
+		// result := make(abi.AbiInterfaceMap)
+		for _, addr := range opts.Addrs {
+			address := common.HexToAddress(addr)
```
tjayrush, line 29: This line reveals a fundamental problem with something. It was always my intention that by the time we were "processing" stuff (that is looping) across any address arrays, they were already lower cased.

In the `options.go` code for every command, we should be doing this so that no-where in any of our code do we ever have to worry. Addresses should always be lower case, and they should be made lower case as soon as we read them in the options code.

We can discuss.

src/apps/chifra/internal/abis/handle_addresses.go
```diff
@@ -0,0 +1,113 @@
...
+			contract, err := account.IsContractAt(
+				opts.Globals.Chain,
+				address,
+				nil, // use latest block number
+			)
+			if err != nil {
+				errorChan <- err
+				cancel()
+			}
+			if !contract {
```
tjayrush, line 67: Couldn't we have done this in the Validate function so we don't even get this far if the user includes non-smart contracts to the tool? (I realize that this is probably how it works in the C++ code, but should it?)

src/apps/chifra/internal/abis/handle_addresses.go
```diff
@@ -0,0 +1,113 @@
...
+				logger.Log(logger.Info, "Address", address, "is not a smart contract. Skipping...")
+				continue
+			}
+			// Fetch ABI from a provider
+			if err = abi.DownloadAbi(opts.Globals.Chain, address, result); err != nil {
+				errorChan <- err
+			}
+			// TODO: This code is duplicated
+			if testMode {
+				// Tests expect sorted map, because maps in C++ are sorted.
```
tjayrush, line 77: I'm not in love with this...we could add a parameter to DownloadAbi and the above function called `sorted` and it could take a value of `testMode`. Pushes this complication down making this code easier to grok.

src/apps/chifra/internal/abis/handle_find.go
```diff
@@ -18,95 +19,123 @@ import (
...
-	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/internal/globals"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/logger"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/output"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/progress"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
 )
 
 func (opts *AbisOptions) HandleAbiFind() error {
+	testMode := opts.Globals.TestMode
```
tjayrush, line 30: I'll just assume this works if the tests pass.

src/apps/chifra/internal/abis/output.go
```diff
@@ -54,12 +54,15 @@ func (opts *AbisOptions) AbisInternal() (err error, handled bool) {
 		return opts.HandleAbiFind(), true
 	}
 
-	if opts.Globals.IsApiMode() {
-		return nil, false
+	if opts.Sol {
```
tjayrush, line 57: Should be `!opts.IsPorted()`

src/apps/chifra/internal/globals/output.go
```diff
@@ -22,7 +22,6 @@ func RenderSlice[
 		types.ReportClean |
 		types.ReportCheck |
 		types.SimpleFunction |
-		types.SimpleFunctionOutput |
```
tjayrush, line 25: Thank you!

src/apps/chifra/pkg/abi/download.go
```diff
@@ -0,0 +1,72 @@
...
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/cache"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+func isZero(address common.Address) bool {
```
tjayrush, line 16: I think there must already be a function like this somewhere. In `utils` maybe?

src/apps/chifra/pkg/account/account.go
```diff
@@ -0,0 +1,35 @@
+package account
```
tjayrush, line 1: I'm not a huge fan of having a package called `account`. I've struggled with this, and the distinction between `account` and `address` gets really fuzzy. Can we put this into the `rpcClient` package? In fact, it may already be in there for other reasons.

src/apps/chifra/pkg/types/types_function.go
```diff
@@ -0,0 +1,185 @@
+package types
+
+import (
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type RawFunction interface{}
```
tjayrush, line 10: This will eventually be auto-generated, just FYI.

src/other/install/abis/known-000/proxy.json
```diff
@@ -1,31 +1,46 @@
 [
   {
     "encoding": "0x073d36b4",
-    "inputs": {
-      "newBeacon": ""
-    },
+    "inputs": [
```
tjayrush, line 4: Nice catch.

test/gold/apps/chifra/chifra_run_abi.txt
```diff
@@ -2,59 +2,59 @@ chifra  abis 0xbb9bc244d798123fde783fcc1c72d3bb8c189413
 TEST[DATE|TIME] Addrs:  [0xbb9bc244d798123fde783fcc1c72d3bb8c189413]
 TEST[DATE|TIME] Format:  txt
 encoding	type	name	signature	input_names	output_names
-0x013cf08b	function	proposals	proposals(uint256)	val_0,	recipient,amount,description,votingDeadline,open,proposalPassed,proposalHash,proposalDeposit,newCurator,yea,nay,creator,,
```
tjayrush, line 5: Weird

test/gold/apps/chifra/chifra_run_abi.txt
```diff
@@ -2,59 +2,59 @@ chifra  abis 0xbb9bc244d798123fde783fcc1c72d3bb8c189413
...
+0x23b872dd	function	transferFrom	transferFrom(address,address,uint256)	_from,_to,_value	success
+0x2632bf20	function	unblockMe	unblockMe()		val_0
+0x34145808	function	totalRewardToken	totalRewardToken()		val_0
+0x39d1f908	function	actualBalance	actualBalance()		_actualBalance
+0x4b6753bc	function	closingTime	closingTime()		val_0
+0x4df6d6cc	function	allowedRecipients	allowedRecipients(address)	val_0	val_0
+0x4e10c3ee	function	transferWithoutReward	transferWithoutReward(address,uint256)	_to,_value	success
+0x5790de2c279e58269b93b12828f56fd5f2bc8ad15e61ce08572585c81a38756f	event	ProposalAdded	ProposalAdded(uint256,address,uint256,bool,string)	proposalID,recipient,amount,newCurator,description	
+0x590e1ae3	function	refund	refund()		
+0x612e45a3	function	newProposal	newProposal(address,uint256,string,bytes,uint256,bool)	_recipient,_amount,_description,_transactionData,_debatingPeriod,_newCurator	_proposalID
```
tjayrush, line 23: Looks good

test/gold/apps/chifra/chifra_run_abi.txt
```diff
@@ -2,59 +2,59 @@ chifra  abis 0xbb9bc244d798123fde783fcc1c72d3bb8c189413
...
+0xbb28353e4598c3b9199101a66e0989549b659a59a54d2c27fbb183f1932c8e6d	event	Refund	Refund(address,uint256)	to,value	
+0xbe7c29c1	function	getNewDAOAddress	getNewDAOAddress(uint256)	_proposalID	_newDAO
+0xc9d27afe	function	vote	vote(uint256,bool)	_proposalID,_supportsProposal	_voteID
+0xcc9ae3f6	function	getMyReward	getMyReward()		_success
+0xcdef91d0	function	rewardToken	rewardToken(address)	val_0	val_0
+0xdbccb92686efceafb9bb7e0394df7f58f71b954061b81afb57109bf247d3d75a	event	CreatedToken	CreatedToken(address,uint256)	to,amount	
+0xdbde1988	function	transferFromWithoutReward	transferFromWithoutReward(address,address,uint256)	_from,_to,_value	success
+0xdd62ed3e	function	allowance	allowance(address,address)	_owner,_spender	remaining
+0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef	event	Transfer	Transfer(address,address,uint256)	_from,_to,_amount	
+0xdfc78bdca8e3e0b18c16c5c99323c6cb9eb5e00afde190b4e7273f5158702b07	event	ProposalTallied	ProposalTallied(uint256,bool,uint256)	proposalID,result,quorum
```
tjayrush, line 54: I wonder if brain-dead third-party code that parses this output will read the missing ending tab as an invalid line. The header suggests X fields in the file, this record (and others with no trailing tab) has X-1 fields.

dszlachta: Hm, the header is on line 4 and has 6 items, all of the lines have 6 items, too. If the last is missing, it's empty, but the tab character is there. So I guess it should read it OK.

src/apps/chifra/pkg/abi/load_abi.go
dszlachta: Yes, I will do it before merging Address thing.

src/apps/chifra/pkg/abi/load_abi.go
dszlachta: Yeah, that'd be probably a better solution. I'll fix that

src/apps/chifra/pkg/abi/load_abi.go
dszlachta: It's called from https://github.com/TrueBlocks/trueblocks-core/blob/develop/src/apps/chifra/pkg/abi/load_abi.go#L88

src/apps/chifra/internal/abis/handle_addresses.go
dszlachta: But we only return if there was an error. Should we ignore it and continue even if we cannot read the binary cache?

src/apps/chifra/internal/abis/handle_addresses.go
dszlachta: Custom Address should solve this

src/apps/chifra/internal/abis/handle_addresses.go
dszlachta: We could, but my understanding of C++ code is that it ignores non-contract addresses, but continues. Putting this into Validate would change this behaviour to "I will not run unless all of the addresses are contracts". We'd also need another loop to check all addresses before running the actual code. But I can be mistaken

src/apps/chifra/internal/abis/handle_addresses.go
dszlachta: Yeah, I hate this code, especially duplication. But I felt that by working longer on it I'm blocking articulation, which blocks whole "port to Go" project, so I rushed a bit. But we can (and should) fix it when articulation is done.

src/apps/chifra/internal/abis/output.go
dszlachta: Good catch 👍

src/apps/chifra/pkg/account/account.go
dszlachta: Sure

src/apps/chifra/pkg/abi/load_abi.go
dszlachta: It comes from go-ethereum directly (they use uppercase), so there is no way to make Encoding lowercase before that code

src/apps/chifra/pkg/account/account.go
dszlachta: I cannot put it in `rpcClient`, because of circular dependencies, but I'll rename it to `contract` instead

PR #2670, opened 2023-01-11
Allows partial ports

tjayrush:


tjayrush:
This allows me to return true or false depending on whether or not various features are ported or not. An example is in `./src/apps/chifra/internal/receipts/output.go` where we return true for everything but `Articulate`.

This makes about 20 of the `chifra blocks` tests to pass in the block porting branch.

Since this code is temporary (it all disappears when we're fully ported), I think that if the tests pass, we can merge it.

dszlachta (approved):

Issue #2667, opened 2023-01-10
chifra transaction data include `finalized` field, but it's confusing


tjayrush:
We hard code the decision of whether or not a transaction is finalized, but it's totally arbitrary.

We should, instead, not export a `finalized` field at all. (This is a breaking change.)

Also, it's confusing as hell because there is now a field in the block called finalized (or at least you can query for a 'finalized' block).

tjayrush:
No longer true of transactions.

PR #2683, opened 2023-01-16
Golang port for chifra blocks 1 version 5

tjayrush:


src/apps/chifra/internal/blocks/handle_count.go
```diff
@@ -0,0 +1,102 @@
...
+		for _, br := range opts.BlockIds {
+			blockNums, err := br.ResolveBlocks(opts.Globals.Chain)
+			if err != nil {
+				errorChan <- err
+				cancel()
+				return
+			}
+
+			for _, bn := range blockNums {
+				finalized := meta.Age(bn) > 28
```
dszlachta, line 38: Do you think it is worth it to add a function, say `isFinalized` and get rid of this free roaming constant?

src/apps/chifra/pkg/rpc/rpc.go
```diff
@@ -0,0 +1,100 @@
...
+
+	theBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(theBytes, ret)
+}
+
+func Query[T any](chain string, method string, params Params) (*T, error) {
```
dszlachta, line 66: This is great 👍

src/apps/chifra/pkg/types/types.go
```diff
@@ -2,24 +2,24 @@ package types
...
 	"github.com/ethereum/go-ethereum/common"
 )
 
 type Modeler[Raw RawData] interface {
-	Model(showHidden bool, format string) Model
+	Model(showHidden bool, format string, extraOptions map[string]any) Model
 	Raw() *Raw
 }
 
+// TODO: Is this needed since we moved to `any`?
```
dszlachta, line 16: Yes, because RawData constrains `Modeler.Raw()` return type and not `Modeler.Model().Model` which is now `any`, but that's same as `interface{}` (so nothing has changed there)

src/apps/chifra/pkg/types/types_appearancecount.go
```diff
@@ -0,0 +1,63 @@
...
+
+// EXISTING_CODE
+
+type RawAppearanceCount struct {
+	Address  string `json:"address"`
+	NRecords string `json:"nRecords"`
+	FileSize string `json:"fileSize"`
+}
+
+type SimpleAppearanceCount struct {
```
dszlachta, line 22: This type is same as `RawAppearanceCount`, are you sure we need the former?

src/apps/chifra/pkg/types/types_namedblock.go
```diff
@@ -11,24 +11,37 @@ package types
...
+	Date        string `json:"date"`
+	Name        string `json:"name"`
+}
 
 type SimpleNamedBlock struct {
 	BlockNumber uint64 `json:"blockNumber"`
-	Timestamp uint64 `json:"timestamp"`
-	Date string `json:"date"`
-	Name string `json:"name,omitempty"`
+	Timestamp   int64  `json:"timestamp"`
```
dszlachta, line 23: This is also same as the Raw type. Maybe we don't need both?

src/apps/chifra/pkg/types/types_transaction.go
```diff
@@ -0,0 +1,187 @@
...
+		model["value"] = s.Value.String()
+		model["receipt"] = nil
+
+		// TODO: these date-related values could be done when RPC is queried and cached
+		model["date"] = date.Format("2006-01-02 15:04:05") + " UTC"
+		model["datesh"] = date.Format("2006-01-02")
+		model["time"] = date.Format("15:04:05") + " UTC"
+
+		if s.Receipt != nil {
+			// TODO: this should be a utility (and used above as well). May be available in go-ethereum. We should check.
```
dszlachta, line 139: I think it's something along the lines of
```
if s.Receipt.ContractAddress == common.HexToAddress("0x0") {
```

src/apps/chifra/pkg/utils/utils.go
```diff
@@ -202,12 +202,25 @@ func GetFields(t *reflect.Type, format string, header bool) (fields []string, se
...
 }
+
+func BigInt_2_Str(bi big.Int) (string, error) {
+	// ret := big.Int{}
+	// if len(str) > 2 && str[:2] == "0x" {
+	// 	ret.SetString(str[2:], 16)
+	// } else {
+	// 	ret.SetString(str, 10)
+	// }
+	return "ret", nil
```
dszlachta, line 225: This looks like it can generate super weird issues, doesn't it?
It could be just `bi.String()` (or we remove this function entirely)

dszlachta (approved):
Looks great 👍

src/apps/chifra/internal/blocks/handle_count.go
tjayrush: Yes. I do. In fact, I think we maybe want to remove it as it's actually confusing and completely arbitrary. I documented it, but I'm not sure what side effects it might have.

https://github.com/TrueBlocks/trueblocks-core/issues/2667

src/apps/chifra/pkg/rpc/rpc.go
tjayrush: It really is, and I was thinking that we could add it to the Modeler interface which would force us to write this routine for each type.

Also -- I woke up thinking the Modeler should have a WriteCache and ReadCache interface as well.

src/apps/chifra/pkg/types/types_appearancecount.go
tjayrush: Probably not, but this is an artifact of the auto-code generation process. I didn't really think about it, but if it was missing (a special case) that special case would flow out into other code I think (like the cache writing/reading code).

src/apps/chifra/pkg/types/types_namedblock.go
tjayrush: Avoiding special cases in auto generated code.

src/apps/chifra/pkg/types/types_transaction.go
tjayrush: I feel like different node software might represent the zero address differently, so we should be permissions.

For example, I've seen `null`, `nil, `0x`, `0x0`, and `0x0000000000000000000000000000000000000000` over the years. Even Erigon itself has (mistakenly) popped back and forth in the past for some fields.

It's probably less of a problem now, but who knows. In the C++ code, there's a function called `isZeroAddress`.

src/apps/chifra/pkg/rpc/rpc.go
dszlachta: Or maybe we need Cacheable interface? Interfaces in Go are really easy to compose, we would just have to add methods to the data type

PR #2704, opened 2023-01-25
Custom Address type returning lowercase Hex

dszlachta:
This Address type always returns lowercase hex


src/apps/chifra/pkg/types/types.go
```diff
@@ -29,6 +30,47 @@ func (h IpfsHash) String() string {
...
+type Address struct {
+	common.Address
+}
+
+// Hex returns string representation of an address
+func (a *Address) Hex() string {
+	if a.IsZero() {
+		return "0x0"
+	}
+	// This is 1000 ns/op faster than strings.ToLower
```
tjayrush, line 44: That's huge when doing 600,000,000 transactions. Just kidding.

PR #2707, opened 2023-01-27
Port articulation part 1

dszlachta:
The first part of porting articulation to Go. It works well, but about 3 test cases still fail. Part two would be to fix these tests and make sure we have a global storage for ABIs that we've already loaded.


src/apps/chifra/internal/transactions/handle_articulate.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+func (opts *TransactionsOptions) HandleArticulate() (err error) {
+	abiMap := make(abi.AbiInterfaceMap)
+	chain := opts.Globals.Chain
+
+	ctx, cancel := context.WithCancel(context.Background())
+	fetchData := func(modelChan chan types.Modeler[types.RawTransaction], errorChan chan error) {
+		for _, rng := range opts.TransactionIds {
+			appearances, err := rng.ResolveTxs(chain)
```
tjayrush: Let's call this `txIds`. I use the word `appearance` is reserved for something very specific (the list of appearances for a particular address). This list is of `tx_ids` that the user gave on the command line.

src/apps/chifra/internal/transactions/handle_articulate.go
```diff
@@ -0,0 +1,119 @@
...
+			if err != nil && !errors.Is(err, ethereum.NotFound) {
+				errorChan <- err
+				cancel()
```
tjayrush: I think in some of our other code (`receipts`?) we don't quit the loop here. We only report the error and continue the loop. (I think. I might be wrong.)

src/apps/chifra/internal/transactions/handle_articulate.go
```diff
@@ -0,0 +1,119 @@
...
+					}
+
+					tx.Traces[index].ArticulatedTrace, err = articulate.ArticulateTrace(&trace, abiMap)
+					if err != nil {
+						errorChan <- err
+						continue
+					}
+				}
+
+				selector := tx.Input[:10]
```
tjayrush: Will this panic if `len(input)` is less than 10?

src/apps/chifra/pkg/articulate/log.go
```diff
@@ -0,0 +1,51 @@
+package articulate
+
+import (
+	"encoding/hex"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/abi"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
+)
+
+func ArticulateLog(log *types.SimpleLog, abiMap abi.AbiInterfaceMap) (articulated *types.SimpleFunction, err error) {
```
tjayrush, line 10: Don't do it now, but at some point, I feel like this will have an interface thus:

```
func (log *types.SimpleLog) Articulate(abiMap abi.AbiInterfaceMap) (articulated *types.SimpleFunction, err error)
```

In other words `logs`, `transactions` and `traces` will implement an interface called `Articulate`.

src/apps/chifra/pkg/articulate/string.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if hasPrintableCharacters {
+		return string(result), true
+	}
+
+	return
+}
+
+func sanitizeByte(character byte) (replacement []byte, replaced int) {
```
tjayrush, line 42: Nice

src/apps/chifra/pkg/articulate/trace.go
```diff
@@ -0,0 +1,51 @@
+package articulate
+
+import (
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/abi"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
+	goEthAbi "github.com/ethereum/go-ethereum/accounts/abi"
+)
+
+func ArticulateTrace(trace *types.SimpleTrace, abiMap abi.AbiInterfaceMap) (articulated *types.SimpleFunction, err error) {
```
tjayrush, line 9: This may one day be `func (trace *types.SimpleTrace) Articulate(abiMap, abi.AbiInterfaceMap)...`

src/apps/chifra/pkg/cache/read.go
```diff
@@ -606,8 +607,11 @@ func ReadParameter(reader *bufio.Reader) (param *types.SimpleParameter, err erro
 		return
 	}
 
-	err = readString(reader, &param.Value)
-	if err != nil {
+	var jsonValue string
```
tjayrush, line 610: Just curious. Does the C++ code store the articulated parameter as a JSON string?

src/apps/chifra/pkg/rpcClient/get_receipt.go
```diff
@@ -106,7 +106,7 @@ func GetTransactionReceipt(chain string, bn uint64, txid uint64, txHash *common.
 	// TODO: are 2 issues with it: 1. circular dependency with types package, 2. every
 	// TODO: call to GetSpecials parses CSV file, so we need to call it once and cache
 	londonBlock := uint64(12965000)
-	if chain == "mainnet" {
+	if tx != nil && chain == "mainnet" {
```
tjayrush, line 110: Can you please add a TODO here: // TODO: chain specific

src/apps/chifra/pkg/rpcClient/get_receipt.go
```diff
@@ -123,17 +123,20 @@ func GetTransactionReceipt(chain string, bn uint64, txid uint64, txHash *common.
...
-		txHashString = txHash.Hex()
-	} else {
-		tx, err = TxFromNumberAndId(chain, bn, txid)
-		if err != nil {
-			return
-		}
-		txHashString = tx.Hash().Hex()
+	// TODO: this optimization makes it hard to know what this function will return.
+	// TODO: A code expecting tx can still supply txHash, in which case it will get nil.
+	// if txHash != nil {
```
tjayrush: Can you remove these comments?

src/apps/chifra/pkg/types/compressed.go
```diff
@@ -0,0 +1,113 @@
...
+		return "{" + content + "}"
+	case reflect.Pointer:
+		return serialize(inputValue.Elem().Interface(), nextMaxDeep)
+	}
+
+	return fmt.Sprint(input)
+}
+
+// MakeCompressed produces compressedTx|Trace|Log field value
+func MakeCompressed(input any) (result string) {
```
tjayrush, line 113: This might one day (not today) turn into 

```
func (tx *types.SimpleTransaction) Compress()...
func (tx *types.SimpleTrace) Compress()...
func (tx *types.SimpleLog) Compress()...
```

so we can remove the `any`.

Please add a TODO: with the above comment.

src/apps/chifra/pkg/types/types_function.go
```diff
@@ -174,12 +287,25 @@ func (s *SimpleFunction) Model(showHidden bool, format string, extraOptions map[
...
 	return Model{
 		Data:  model,
 		Order: order,
 	}
 }
+
+// Normalize sets StateMutability from `payable` field. It is only useful when
+// reading ABIs generated before Solidity 0.5.0, which use `payable` field:
+// https://docs.soliditylang.org/en/develop/050-breaking-changes.html#command-line-and-json-interfaces
+func (s *SimpleFunction) Normalize() {
```
tjayrush, line 335: Nice

src/apps/chifra/pkg/types/types_function_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+)
+
+func TestFunctionToAbiMethod(t *testing.T) {
+	input := &SimpleFunction{
```
tjayrush, line 11: Thanks for taking the time to write tests. I appreciate it a lot. Makes me feel comfortable.

src/apps/chifra/pkg/types/types_parameter.go
```diff
@@ -4,10 +4,10 @@ import "fmt"
 
 type RawParameter interface{}
 type SimpleParameter struct {
-	ParameterType string            `json:"parameterType"`
```
tjayrush: I made a change in the auto code generation that might put this back to `type`. No big deal, just FYI.

src/apps/chifra/pkg/types/types_test.go
```diff
@@ -50,6 +51,17 @@ func TestAddress_Hex(t *testing.T) {
 	}
 }
 
+func TestAddress_Stringer(t *testing.T) {
```
tjayrush, line 60: Thanks!

src/apps/chifra/pkg/types/types_trace.go
```diff
@@ -66,15 +66,26 @@ func (s *SimpleTrace) Model(showHidden bool, format string, extraOptions map[str
 	// EXISTING_CODE
 	// EXISTING_CODE
 
-	model := map[string]interface{}{
```
tjayrush, line 69: The auto code generation wants to assign the first version and then overwrite it, but that's easy to fix. Nothing to do here, but you'll see how this works when I merge my PR for makeClass.

src/apps/chifra/pkg/types/types_trace.go
```diff
@@ -286,6 +310,9 @@ func GetTracesByTransactionHash(chain string, txHash string) ([]SimpleTrace, err
 				Action:           &action,
 				Result:           result,
 			}
+			if transaction != nil {
```
tjayrush, line 303: It's very, very unclear to me why the RPC doesn't do this already. It's so annoying.

src/apps/chifra/pkg/types/types_traceaction.go
```diff
@@ -62,7 +62,7 @@ func (s *SimpleTraceAction) Model(showHidden bool, format string, extraOptions m
 		if s.Value.String() != "0" {
 			model["value"] = s.Value.String()
 		}
-		if len(s.Init) > 0 {
+		if extraOptions["tracesTransactionsFormat"] != true && len(s.Init) > 0 {
```
tjayrush: I'm a bit confused by this. In general, when there's a command line option to a tool (such as `chifra transactions --traces`) I would have used `traces` instead of `tracesTransactionsFormat`. It just causes less cognitive load since there is no "new word" to understand. I will probably change this once it's merged.

src/apps/chifra/pkg/types/types_transaction.go
```diff
@@ -172,4 +233,9 @@ func (s *SimpleTransaction) Model(showHidden bool, format string, extraOptions m
 }
```
tjayrush, line 269: I'm just noting here that the "rendering" code can get quite complicated given all the options. At some point (not now), we may revisit how some of these things work. For example, the many alternative "formats" might be "raised up" into the calling code so that this "type-related" code just does as instructed. We talked about this.

src/other/install/abis/known-000/proxy.json
```diff
@@ -98,7 +98,8 @@
     "outputs": [
       {
         "name": "storage r",
-        "type": "struct StorageSlot.BooleanSlot"
```
tjayrush, line 101: Nice. I'm curious though. Doesn't it require us to have the structure definition in order to articulate this? If I understand this, it doesn't allow the code to decode the bytes any better than existing. Maybe it's here because otherwise the ABI parser fails?

src/other/install/abis/known-005/erc_01594.json
```diff
@@ -143,7 +143,7 @@
       },
       {
         "name": "ret_1",
-        "type": "byte"
+        "type": "bytes1"
```
tjayrush, line 146: Very strange. How did you find this? What is this? Is a byte and a bytes1 different things?

test/gold/tools/getTrans/getTrans_articulate.txt
```diff
@@ -37,7 +37,7 @@ params.size: 0
         "subtraces": 0,
         "traceAddress": null,
         "type": "call",
-        "error": "invalid jump destination",
```
tjayrush, line 40: This is very annoying. Erigon changed their error message in the latest release. One of my machines (linux) has the updated Erigon. The other (my laptop) does not.

tjayrush (commented):
Generally, it looks very good. All of it makes sense.

There's a few comments in here where I'm saying something like, "Don't fix it now, but put a TODO list item..." It's just a way for me to note for you my thinking and so I can find it later when we can focus on other things.

src/apps/chifra/internal/transactions/handle_articulate.go
dszlachta: I've renamed it, but it is a list of appearances :) 
```golang
func (id *Identifier) ResolveTxs(chain string) ([]types.SimpleAppearance, error) {
// ...
}
```

src/apps/chifra/internal/transactions/handle_articulate.go
dszlachta: Good catch.

src/apps/chifra/pkg/cache/read.go
dszlachta: Not really, but it stores `Value` as string, always. 
This is a shortcut to read the value as string and then convert to the proper type. This makes sure that we output cached parameter's value just like non-cached one (the new code allows `Value` to be anything to make arrays first-class citizens when outputting JSON).

And, because `"some string value"` is valid JSON, it doesn't break when some simple type is cached

src/apps/chifra/pkg/rpcClient/get_receipt.go
dszlachta: Added 👍

src/apps/chifra/pkg/rpcClient/get_receipt.go
dszlachta: Sure, but I think this is an important issue. Whole `rpcClient` package doesn't support the idea of "I already have some parts of the data", as far as I know. Maybe we could add an issue.

src/apps/chifra/pkg/types/types_function_test.go
dszlachta: It makes my anxiety go away 😄

src/apps/chifra/pkg/types/types_traceaction.go
dszlachta: I understand, but the thing is that `chifra transactions --articulate` *always* wants different format of traces. So we need something in `SimpleTrace.Model()` to tell if it's being run by `chifra transactions`... If we use just `traces = opts.Traces`, then we don't know which command is calling this code. But we can change the name to something simpler. It's also super confusing that different tools want different fields of the same data, I don't know if we can simplify it, but if we can then maybe we should

src/other/install/abis/known-000/proxy.json
dszlachta: Yeah, the only reason is that go-ethereum's parser fails. It also won't articulate because the type should be `(key1Type,key2Type,...)` like `(uint8,bytes32)`. I'm not sure C++ was articulating this? Or just presenting the type to the user?

src/other/install/abis/known-005/erc_01594.json
dszlachta: The parser failed. These are the same thing, but `byte` has been removed in Solidity 8 in favour of `bytes1`: https://docs.soliditylang.org/en/latest/080-breaking-changes.html#silent-changes-of-the-semantics (last bullet point)

test/gold/tools/getTrans/getTrans_articulate.txt
dszlachta: I've seen you change the tests in your PR. I guess this will fix itself when we merge

dszlachta:
Okay, the only failing tests are related to the removal of `input|output_names` and Erigon changing their error message.
I checked out back known ABI files (from develop), because I was worried that C++ could break (I didn't see it in the tests, but still). I've put the files that I fixed into `src/other/install/abis/standard` (I couldn't come up with a better name)

src/apps/chifra/internal/transactions/handle_articulate.go
tjayrush: I see. I guess I can expand my definition of an appearance to "An appearance is an Appearance, but only when used in `chifra export`."

src/apps/chifra/pkg/cache/read.go
tjayrush: Isn't this reading binary code? If the JSON object is not stored as a string, wouldn't this crash? Just curious. Maybe I'm misunderstanding.

dszlachta:
This is how it works:
Some data type —> (JSON encoder) —> string —> (cache/write.go) —> bytes
So, the code here reads the bytes, converts them to string, then parses this string as JSON and we get the original data type

Dawid Szlachta

src/apps/chifra/internal/transactions/handle_articulate.go
tjayrush: Did you see this?

src/apps/chifra/pkg/types/types_traceaction.go
tjayrush: Yes. This whole thing is confusing. I'm certain we're going to revisit the model code. One thing I've had in mind is that each different "version" of the data should actually have its own model.

src/apps/chifra/internal/transactions/handle_articulate.go
dszlachta: I haven't, sorry! Yes, it'd panic. Good catch.

test/gold/tools/getTrans/getTrans_articulate.txt
tjayrush: We have to update our endpoints. That's easy for me, but now that I think about it, it may not be all that easy for you on dAppNode. (This is another reason for us to have our own package.)

tjayrush:
> Okay, the only failing tests are related to the removal of `input|output_names` and Erigon changing their error message. I checked out back known ABI files (from develop), because I was worried that C++ could break (I didn't see it in the tests, but still). I've put the files that I fixed into `src/other/install/abis/standard` (I couldn't come up with a better name)

I'm going to go ahead and merge this and then copy these`standard` (repaired) ones back in. As long as it doesn't break tests, it's better. It can't be worse because the C++ code is ridiculous.

tjayrush:
All tests pass locally. As soon as they pass on the CI, we merge. I'll worry about the other branch separately. I think it's easier this way. I'll probably close that other PR.

tjayrush:
Let's both read and discuss this.

https://www.reddit.com/r/golang/comments/112vfr3/how_to_deal_with_java_developers_polluting_the_go/

I think it's great to consider that "Go is not object oriented," but it can go too far. The article mentions a number of things one shouldn't do (such as Contexts and Channels in a type), but I feel like this attitude can go too far. In some cases, it throws out the baby with the bathwater.

This is showing up in at least three places: (1) `cache` package, (2) `articulation` package, (3) `rpcClient` package. In each of these three packages, we have functions that deal with types from the `types` package. This is apparent given that the functions start with a pointer to the type.

The effect of this is to fragment the type, spreading it across disparate packages. This makes it more difficult to understand the type, change the type, and very importantly for me auto-generate the type, and even more importantly for me, use the type.

An example of 'use the type' is the fact that I want to use the code in `./src/apps/internal/transations/output.go` that articulates the logs (or traces) but in order to do so, I need to copy and paste that code, whereas if it were a function on the type, I could just call it.

We can discuss in Denver.

tjayrush:
NEVER MIND. I WAS RUNNING AGAINST A BOGUS NODE RPC ENDPOINT

~@dawid Two things, please. 1) Can you please merge develop into this PR? I tried, but I feel like I was messing it up. 2) When I do: ` make -j 12 && cd ../src/apps/chifra/ ; go test --tags integration ./... ; cd - ` It reports errors. Can you fix those?~

tjayrush:
~@dawid Sorry if I'm mistaking something. Can you please re-base this to the `develop` branch when you have a chance?~

Actually. Never mind. I was able to do it. I'll include the merge to `develop` in my PR that fixes tests. Coming in a minute.

PR #2728, opened 2023-02-16
Re-enables tests for chifra abis --known

tjayrush:
Work that will finally fully port `chifra abis` by finishing `chifra abis --known` which had been previously untested and is not working.


tjayrush:
Fully removes all C++ code for `chifra abis`.

PR #2731, opened 2023-02-18
Feature/better names

tjayrush:
Implements `chifra names --tags` and `chifra names --addr` in GoLang


src/apps/chifra/pkg/names/names.go
```diff
@@ -161,27 +121,29 @@ type NameReader struct {
 	csvReader csv.Reader
 }
 
-func (gr *NameReader) Read() (Name, error) {
+func (gr *NameReader) Read() (types.SimpleName, error) {
```
dszlachta, line 124: If it's `NameReader`, why is it called `gr`? :)

dszlachta (approved):

src/apps/chifra/pkg/names/names.go
tjayrush: Came from old GitCoin code, so `gr` stands for `grant`. Fixed.

tjayrush:
I'm going to merge this, but I'll follow up quickly with another PR that removes the `SimplePart` code you mentioned. I was able to figure out how to (VERY EASILY) use GoLang templates.

PR #2735, opened 2023-02-19
Draft: `chifra abis`: check if contract is proxy

### LINKED ISSUE #2711: chifra abis - doesn't appear to know about proxies
This command

```
chifra abis 0x99b36fdbc582d113af36a21eba06bfeab7b9be12
```

reports nothing. But this command

```
chifra explore 0x99b36fdbc582d113af36a21eba06bfeab7b9be12
```

https://etherscan.io/address/0x99b36fdbc582d113af36a21eba06bfeab7b9be12#readProxyContract

shows a smart contract that is a proxy. If EtherScan can figure this out, we should be able to as well.

CyMule:
There is still some work to do, but I figured I should create a draft so I can get feedback as I go.

The current work does checks to see if a contract is a proxy (only if it matches EIP1967 at the moment), and it will concatenate the ABIs.

I need to make the implementation ABI show up differently under the following block:
```
encoding        type    name    signature       input_names     output_names
```
where type might say `proxied-function` or `proxied-event` as discussed in #2711

I also still need to add some testing.


tjayrush:
Thanks @CyMule 

@dawid Can you give this a quick review (and welcome CyMule)?

I have some thoughts on how I would review it, but I just want to see your thoughts first.

I'll merge it tomorrow morning (Monday morning).

src/apps/chifra/internal/abis/handle_addresses.go
```diff
@@ -22,28 +23,56 @@ func (opts *AbisOptions) HandleAddresses() (err error) {
...
-					continue
-				} else {
-					// It's okay to not find the ABI. We report an error, but do not stop processing
-					if err = abi.DownloadAbi(opts.Globals.Chain, address, result); err != nil {
-						errorChan <- err
-					}
-				}
+			err := abi.LoadAbiFromAddress(opts.Globals.Chain, address, result)
+			// If there is no error the LoadAbiFromAddress was successful so the ABI is in the result
+			if err == nil {
```
dszlachta, line 29: This looks way clearer now, thanks 👍

src/apps/chifra/internal/abis/handle_addresses.go
```diff
@@ -22,28 +23,56 @@ func (opts *AbisOptions) HandleAddresses() (err error) {
...
+			if contract, err := contract.IsContractAt(opts.Globals.Chain, address, nil); err != nil {
+				errorChan <- err
+				cancel()
+			} else if !contract {
+				logger.Log(logger.Info, "Address", address, "is not a smart contract. Skipping...")
+				continue
+			}
+
+			proxy, err := contract.IsProxy(opts.Globals.Chain, address, nil)
+			fmt.Println("Address", address, "is a proxy:", proxy)
```
dszlachta: If this is only for debugging, let's make sure to remove it before merging

src/apps/chifra/internal/abis/handle_addresses.go
```diff
@@ -22,28 +23,56 @@ func (opts *AbisOptions) HandleAddresses() (err error) {
...
+			// An unexpected error occurred
+			if err != nil {
+				errorChan <- err
+				cancel()
+			}
+
+			if proxy {
+                proxyAddress, err := contract.GetProxyAddress(opts.Globals.Chain, address, nil)
```
dszlachta: Maybe we could fix indentation here before merging

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,62 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+	defer client.Close()
+
+	var clientBlockArg *big.Int = nil
+	if block != nil && block.Name != "latest" {
+		clientBlockArg = big.NewInt(0).SetUint64(block.BlockNumber)
+	}
+	// EIP 1967: https://eips.ethereum.org/EIPS/eip-1967
+	// bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1) == 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc
+	eip1956Position := common.HexToHash("0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc")
+	data, err := client.StorageAt(
```
dszlachta: It seems like we're not checking the `err`

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,62 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+		context.Background(),
+		common.HexToAddress(address.Hex()),
+		eip1956Position,
+		clientBlockArg,
+	)
+
+	// convert []byte to hex string, get the last 20 bytes, and convert to address
+	proxyAddress := common.HexToAddress(fmt.Sprintf("0x%x", data[len(data)-20:]))
```
dszlachta: Can `data` length be less than 20 bytes? If yes, then we should check it first, otherwise we will get a panic

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,62 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+	// address := types.HexToAddress(addr)
+    proxyString := proxyAddress.Hex()
+    address = types.HexToAddress(proxyString)
```
dszlachta: I think that you could replace the above 2 lines with just this one (line 63). 
If I understand correctly, what happens here is: turn bytes to `common.Address`, then convert `common.Address` to `string`, then convert `string` to `types.Address`. `types.HexToAddress` turns a string into `types.Address`, so we can take a shortcut here

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,62 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+   // convert proxtAddress to types.Addres
+	fmt.Println("proxyAddress", proxyAddress.Hex())
```
dszlachta: Debug print?

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,62 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+	defer client.Close()
+
+	var clientBlockArg *big.Int = nil
+	if block != nil && block.Name != "latest" {
+		clientBlockArg = big.NewInt(0).SetUint64(block.BlockNumber)
+	}
+	// EIP 1967: https://eips.ethereum.org/EIPS/eip-1967
+	// bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1) == 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc
+	eip1956Position := common.HexToHash("0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc")
+	data, err := client.StorageAt(
```
dszlachta: Should we check `err`?

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,62 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+		common.HexToAddress(address.Hex()),
+		eip1956Position,
+		clientBlockArg,
+	)
+
+	// convert []byte to hex string, get the last 20 bytes, and convert to address
+    tmpAddress := common.HexToAddress(fmt.Sprintf("0x%x", data[len(data)-20:]))
+	// address := types.HexToAddress(addr)
+    proxyString := tmpAddress.Hex()
+    proxyAddress = types.HexToAddress(proxyString)
```
dszlachta: Same comments about `data` length and `types.Address` as in `IsProxy()`

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,62 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+	// address := types.HexToAddress(addr)
+    proxyString := proxyAddress.Hex()
+    address = types.HexToAddress(proxyString)
+   // convert proxtAddress to types.Addres
+	fmt.Println("proxyAddress", proxyAddress.Hex())
+	proxy, err = IsContractAt(chain, address, block)
+	return
+}
+
+func GetProxyAddress(chain string, address types.Address, block *types.SimpleNamedBlock) (proxyAddress types.Address, err error) {
```
dszlachta: It seems to me that this function is same as `IsProxy()`, but returns an address instead of `boolean`. Do you think we could get rid of the duplicated code (and possibly duplicated execution) somehow? For example, maybe `GetProxyAddress` could return a pointer to address and if it's nil then we'd know that the contract is not a proxy?

dszlachta (changes requested):
Thanks @CyMule, it looks great, but I had some questions and also added a few suggestions.

CyMule:
@dszlachta thank you for the feedback! I made the requested changes and cleaned up the code a bit more.

The code also checks for the known implementation address locations listed [here](https://github.com/TrueBlocks/trueblocks-core/blob/master/src/libs/etherlib/ethcall.cpp#L534-L540).

I still need to:
- Add a proxy check in `LoadAbiFromAddress`. Otherwise the second call to `chifra abis $ADDRESS` only returns records for `$ADDRESS` and doesn't check the cache for the discovered implementation address
- Change the display of the implementation contract's ABI to show `proxied-function` or `proxied-event`. I am not 100% certain the best way to do this yet. I believe I will do this by storing the ABI for the implementation in its own `abi.AbiInterfaceMap` and find a good way to pass the two separate `abi.AbiInterfaceMap`'s through `modelChan' and indicate that one is for the proxy and one is for the implementation
- Add tests for the new functionality

Feel free to review the progress now or wait until I have completed these additional steps.

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,72 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+	for _, slot := range slots {
+		slotStorage, err := client.StorageAt(
+			context.Background(),
+			common.HexToAddress(address.Hex()),
+			slot,
+			clientBlockArg,
+		)
+		// In theory all storage slots are 32 bytes, but we should check in case
+		if err != nil || len(slotStorage) < 20 {
+			break
```
dszlachta, line 92: It seems like this code won't report the error. You're creating a new variable named `err` in scope of this `for` loop and the value is not being assigned to the top-level `err` that gets returned on line 101 and 104. Here's an example: https://go.dev/play/p/rg3DpavcLWc

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,72 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+		clientBlockArg = big.NewInt(0).SetUint64(block.BlockNumber)
+	}
+	// EIP 1967: https://eips.ethereum.org/EIPS/eip-1967
+	// bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1) == 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc
+	eip1967Slot := common.HexToHash("0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc")
+	eip1967ZOSSlot := common.HexToHash("0x7050c9e0f4ca769c69bd3a8ef740bc37934f8e2c036e5a723fd8ee048ed3f8c3")
+	eip1822Slot := common.HexToHash("0xc5f16f0fcc639fa48a6947836d9850f504798523bf8c9a3a87d5876cf622bcf7")
+	eip1822ZOSSlot := common.HexToHash("0x5f3b5dfeb7b28cdbd7faba78963ee202a494e2a2cc8c9978d5e30d2aebb8c197")
+	// This is the slot used by the OpenZeppelin GnosisSafeProxy
+	zeroSlot := common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000000000")
```
dszlachta, line 80: A shorter way to get the same value is `common.HexToHash("0x0")`

src/apps/chifra/pkg/contract/contract.go
```diff
@@ -33,3 +34,72 @@ func IsContractAt(chain string, address types.Address, block *types.SimpleNamedB
...
+	potentialAddresses, err := getKnownImplementationSlots(chain, address, block)
+	if err != nil {
+		return
+	}
+
+	// check if any of the potential addresses are contracts
+	// if any are contracts, address is a proxy and potentialAddress is the implementation
+	for _, potentialAddress := range potentialAddresses {
+		isContract, err := IsContractAt(chain, potentialAddress, block)
+		if err != nil {
```
dszlachta, line 50: Same comment about `err` as here https://github.com/TrueBlocks/trueblocks-core/pull/2735/files#r1112935936

PR #2752, opened 2023-03-05
Better/make class2

tjayrush:


tjayrush:
This PR contains mostly formatting and auto-code generation cleanups. Some changes to types, but non breaking (other than one to TraceReceipt to align the name of the old `newContract` field to `address` which aligns with Erigon RPC.

PR #2765, opened 2023-03-09
Feature/port articulation fix tests

dszlachta:
Corrected all tests but `hex_2_string_` (@tjayrush could you have a look? In working there are `gasPrice` and `gasCost` columns, in gold `ethGasPrice` and no gasCost - I don't know which ones we want to show) and `revert_no_error` (it's fixed, but go-ethereum adds number when method of the same name is already present - I don't know if that's OK).

Also `issue_1583` is not fixed yet, because I can't get it to load correct ABI definition and it tries to load wrong input types.


tjayrush:
> In working there are `gasPrice` and `gasCost` columns, in gold `ethGasPrice` and no gasCost - I don't know which ones we want to show)

I wanted to remove `ethGasPrice` and replace it with `gasPrice` and add a new column `gasCost`.

tjayrush:
> it's fixed, but go-ethereum adds number when method of the same name is already present - I don't know if that's OK

I was wondering where that was coming from. If it comes from Geth, then I guess it's going to have to be okay. I can't see a good reason to post-process it (nor a good way to do so).

tjayrush:
I'm going to go ahead and merge this into the underlying PR and then work out remaining issues with a PR against that.

dszlachta:
> > it's fixed, but go-ethereum adds number when method of the same name is already present - I don't know if that's OK
> 
> I was wondering where that was coming from. If it comes from Geth, then I guess it's going to have to be okay. I can't see a good reason to post-process it (nor a good way to do so).

Just FYI, it comes from `go-ethereum` package (so saying it comes from Geth should be true, too). They explain it in the first comment about Name here: https://pkg.go.dev/github.com/ethereum/go-ethereum/accounts/abi#Method

tjayrush:
Is there anything short of a total hack that would allow us to remove it? It's almost certainly related to the issue wherein the same function with some of its parameters indexed and some not indexed brings the same four-byte. We need to solve that problem as well.

PR #2812, opened 2023-03-24
Allows starting the scraper from chifra daemon

tjayrush:


tjayrush:
Closed because of _2.

PR #2856, opened 2023-04-13
gRPC names server

dszlachta:
This PR adds gRPC server to `chifra daemon` and modifies `chifra names` so that it fetches the data through gRPC if the server is running. Currently the server only works with Unix Domain Sockets (super fast, but local connections only).


tjayrush:
@dawid. When I try to build this on my machine, I get this:

```
-- Cleaning old files...
[  0%] Built target init_configs
[  2%] Built target acct
[ 16%] Built target ether
protoc-gen-go-grpc: program not found or is not executable
Please specify a program using absolute path or make sure the program is available in your PATH system variable
--go-grpc_out: protoc-gen-go-grpc: Plugin failed with status code 1.
make[2]: *** [chifra_proto] Error 1
make[1]: *** [apps/chifra/CMakeFiles/chifra_proto.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 33%] Built target util
[ 33%] Built target chifra
```

I guess there's two things we can do:

1. Figure out if CMake is able to download and install the protoc tool (so the user doesn't have to), or
2. Write a short entry in the installation website (and possibly the README) to help the user figure out what's required to build.

There's instructions here for downloading build requirements (https://trueblocks.io/docs/install/install-troubleshooting/#installing-build-tools). Maybe we can add something there or somewhere near here: https://trueblocks.io/docs/install/install-core/#installation.

If it's only a path problem, I think CMake should be able to find it. Search this file (https://github.com/TrueBlocks/trueblocks-core/blob/master/src/CMakeLists.txt#L41) which finds Curl or fails to build with a better message.

tjayrush:
On a mac, it's `brew install protoc-gen-go-grpc`. Not sure what it should be on Linux.

tjayrush:
I did a very simple test. I made a file called `test` with 100 copies of this line:

```
chifra names >/dev/null
```

and ran it with `time source test` against the daemon WITHOUT `--rpc`.  `24.2` seconds.

I ran the same test against the daemon with `--rcp`. 26 seconds.

I then changed `chifra names` to `curl http://localhost/name` and ran it again: 30 seconds.

Maybe there's something weird about my test, but I would have expected very significant speedup. Are you sure we're caching in memory?

Is there a better way to test this?

tjayrush:
Sorry to keep piling on, but when I run `make generate finish`, it breaks the build. I think the entry you have in the `cmd-line-options.csv` file should be a boolean not an enum. (You can find examples of a boolean in other commands.)

Also, I wonder if we can make `rpc` say `grpc` throughout. It's confusing to use `rpc` in some places and `grpc` in others. If you change the line in the `cmd-line-options.csv` file, and re-run `make generate finish`, it will regenerate all the code and the help text.

I would normally do this for you, but I guess it doesn't hurt to have you do it. Running `make generate finish` should be part of posting any PR that changes options, I think.

dszlachta:
> Sorry to keep piling on, but when I run `make generate finish`, it breaks the build. I think the entry you have in the `cmd-line-options.csv` file should be a boolean not an enum. (You can find examples of a boolean in other commands.)
> 
> Also, I wonder if we can make `rpc` say `grpc` throughout. It's confusing to use `rpc` in some places and `grpc` in others. If you change the line in the `cmd-line-options.csv` file, and re-run `make generate finish`, it will regenerate all the code and the help text.
> 
> I would normally do this for you, but I guess it doesn't hurt to have you do it. Running `make generate finish` should be part of posting any PR that changes options, I think.

Sorry, it's fixed now

dszlachta:
> I did a very simple test. I made a file called `test` with 100 copies of this line:
> 
> ```
> chifra names >/dev/null
> ```
> 
> and ran it with `time source test` against the daemon WITHOUT `--rpc`. `24.2` seconds.
> 
> I ran the same test against the daemon with `--rcp`. 26 seconds.
> 
> I then changed `chifra names` to `curl http://localhost/name` and ran it again: 30 seconds.
> 
> Maybe there's something weird about my test, but I would have expected very significant speedup. Are you sure we're caching in memory?
> 
> Is there a better way to test this?

For future reference: after investigating with profiler, it seems that the bottleneck here are the syscalls required to print stuff to the screen. It seems like our code in both cases is faster than syscalls, that is why we see no significant difference in execution time between stand-alone and gRPC-connected runs. Maybe the difference will be visible when we add gRPC support for other features, like ABIs

dszlachta:
> When I try to build this on my machine, I get this:
> 
> ```
> -- Cleaning old files...
> [  0%] Built target init_configs
> [  2%] Built target acct
> [ 16%] Built target ether
> protoc-gen-go-grpc: program not found or is not executable
> Please specify a program using absolute path or make sure the program is available in your PATH system variable
> --go-grpc_out: protoc-gen-go-grpc: Plugin failed with status code 1.
> make[2]: *** [chifra_proto] Error 1
> make[1]: *** [apps/chifra/CMakeFiles/chifra_proto.dir/all] Error 2
> make[1]: *** Waiting for unfinished jobs....
> [ 33%] Built target util
> [ 33%] Built target chifra
> ```
> 
> I guess there's two things we can do:
> 
>     1. Figure out if CMake is able to download and install the protoc tool (so the user doesn't have to), or
> 
>     2. Write a short entry in the installation website (and possibly the README) to help the user figure out what's required to build.
> 
> 
> There's instructions here for downloading build requirements (https://trueblocks.io/docs/install/install-troubleshooting/#installing-build-tools). Maybe we can add something there or somewhere near here: https://trueblocks.io/docs/install/install-core/#installation.
> 
> If it's only a path problem, I think CMake should be able to find it. Search this file (https://github.com/TrueBlocks/trueblocks-core/blob/master/src/CMakeLists.txt#L41) which finds Curl or fails to build with a better message.

This is fixed: cmake will try to generate gRPC code only if the user has `protoc` installed. Users should not need to generate it, only we have to do it.

docs/readmes/admin-daemon.md
```diff
@@ -29,6 +29,7 @@ Flags:
   -s, --scrape string   start the scraper, initialize it with either just blooms or entire index, generate for new blocks
                         One of [ off | blooms | index ]
   -m, --monitor         instruct the node to start the monitors tool
+  -r, --rpc             run gRPC server
```
tjayrush: Change this to `-g, --grpc       run gRPC server to serve names`

Do this in the `./src/cmd-line-options.csv` file and run `make generate finish` to update all the help files and source code.

This will break the one line where you reference `opts.RPC`, but other than that it should be minimal.

src/apps/chifra/cmd/daemon.go
```diff
@@ -59,6 +59,7 @@ One of [ off | on ]`)
 	daemonCmd.Flags().StringVarP(&daemonPkg.GetOptions().Scrape, "scrape", "s", "", `start the scraper, initialize it with either just blooms or entire index, generate for new blocks
 One of [ off | blooms | index ]`)
 	daemonCmd.Flags().BoolVarP(&daemonPkg.GetOptions().Monitor, "monitor", "m", false, "instruct the node to start the monitors tool")
+	daemonCmd.Flags().BoolVarP(&daemonPkg.GetOptions().Rpc, "rpc", "r", false, "run gRPC server")
```
tjayrush: This will change when you re-run `make generate finish`.

src/apps/chifra/internal/daemon/handle_rpc.go
```diff
@@ -0,0 +1,144 @@
+package daemonPkg
```
tjayrush: Rename this file to `handle_grpc.go`, please.

src/apps/chifra/internal/daemon/handle_rpc.go
```diff
@@ -0,0 +1,144 @@
...
+	} else {
+		_, err = names.RemoveCustomName(request.Chain, addr)
+	}
+	return &proto.CRUDResponse{
+		Success: err == nil,
+		Error:   utils.PointerOf(err.Error()),
+	}, err
+}
+
+func (opts *DaemonOptions) HandleRpc() error {
```
tjayrush: HandleGrpc

src/apps/chifra/internal/daemon/handle_rpc_integration_test.go
```diff
@@ -0,0 +1,66 @@
+package daemonPkg
+
+import (
+	"context"
+	"net"
+	"testing"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/proto"
+	grpc "google.golang.org/grpc"
```
tjayrush: Just curious. Is this actually required? It's not in the other file above.

src/apps/chifra/internal/names/handle_terms.go
```diff
@@ -2,37 +2,93 @@ package namesPkg
...
-	// Note: Make sure to add an entry to enabledForCmd in src/apps/chifra/pkg/output/helpers.go
-	fetchData := func(modelChan chan types.Modeler[types.RawName], errorChan chan error) {
-		for _, name := range namesArray {
-			name := name
-			modelChan <- &name
+		namesArray, err := names.LoadNamesArray(opts.Globals.Chain, opts.getType(), names.SortByAddress, opts.Terms)
+		if err != nil {
+			return err
+		}
+		if len(namesArray) == 0 {
```
tjayrush, line 38: Do we handle the `no results` case in the RPC server path?

src/apps/chifra/internal/names/handle_terms.go
```diff
@@ -2,37 +2,93 @@ package namesPkg
...
+			logger.Warn("No results for", os.Args)
+			return nil
 		}
-	}
 
+		// Note: Make sure to add an entry to enabledForCmd in src/apps/chifra/pkg/output/helpers.go
```
tjayrush: This comment can be removed. It's no longer true

src/apps/chifra/internal/names/handle_terms.go
```diff
@@ -2,37 +2,93 @@ package namesPkg
...
 	extra := map[string]interface{}{
 		"expand":  opts.Expand,
 		"prefund": opts.Prefund,
 	}
-	return output.StreamMany(ctx, fetchData, opts.Globals.OutputOptsWithExtra(extra))
+	if opts.Addr {
+		extra["single"] = "address"
+	}
+	ctx := context.Background()
+	return output.StreamMany(ctx, fetchData, output.OutputOptions{
```
tjayrush, line 58: This can be collapsed. See other Handle routines throughout the code such as here: https://github.com/TrueBlocks/trueblocks-core/blob/master/src/apps/chifra/internal/blocks/handle_trace.go#L61

src/apps/chifra/proto/chifra.proto
```diff
@@ -0,0 +1,55 @@
...
+    optional bool isContract = 4;
+    optional bool isCustom = 5;
+    optional bool isErc20 = 6;
+    optional bool isErc721 = 7;
+    optional bool isPrefund = 8;
+    string name = 9;
+    optional string petname = 10;
+    optional string source = 11;
+    optional string symbol = 12;
+    optional string tags = 13;
```
tjayrush, line 39: I'm not quite sure what `optional` means here, but I do know that `tags` cannot be empty. I think `clean` (which should be called by all the `create` and `edit` crud routines should default it to `30-Contracts` if it's a contract and something else if it's not a contract. I'm pretty sure we have tests for that, but if not, we should.

src/cmd-line-options.csv
```diff
@@ -228,6 +228,7 @@ num,group,tags,api_route,tool,longName,hotKey,def_val,is_required,is_customizabl
 13114,apps,Admin,daemon,flame,api,a,on,false,false,true,false,gocmd,flag,enum[off|on*]>,instruct the node to start the API server
 13112,apps,Admin,daemon,flame,scrape,s,,false,false,true,false,gocmd,flag,enum[off|blooms|index]>,start the scraper&#44; initialize it with either just blooms or entire index&#44; generate for new blocks
 13113,apps,Admin,daemon,flame,monitor,m,,false,false,true,false,gocmd,switch,<boolean>,instruct the node to start the monitors tool
+13118,apps,Admin,daemon,flame,rpc,r,,false,false,true,true,gocmd,switch,<boolean>,run gRPC server
```
tjayrush: change this to

```
...,grpc,g,....run gRPC server to serve names
```

and then run `make generate finish`.

src/apps/chifra/internal/daemon/handle_rpc_integration_test.go
dszlachta: It wasn't, good catch

src/apps/chifra/internal/names/handle_terms.go
dszlachta: Added, thanks 👍

src/apps/chifra/internal/names/handle_terms.go
dszlachta: I saw it and thought about the same thing earlier. But here we need to override `NoHeader` if `opts.Addr` is true

src/apps/chifra/proto/chifra.proto
dszlachta: Optional means that you don't have to set this field. I can make `tags` required, but from what you write it's optional, because even if it's missing, it gets set by `clean`.

PR #2859, opened 2023-04-15
Develop

tjayrush:


tjayrush:
Fixes a regression due to not updating go-ethereum dependancy

PR #2860, opened 2023-04-15
Implement names --clean

dszlachta:
Implements `chifra names --clean`.
Adds 2 new `chifra names` options: 
1. `--dry_run` (output changes to stdout instead of writing files)
2. `--regular` (clean regular database, the default is to clean custom names)

Also adds `utils.IterateOverMap[Key, Value]()`, which allows us to process map items in batches concurrently (90% faster than synchronous Go version).

We may want to squash merge it.


src/apps/chifra/internal/names/handle_clean.go
```diff
@@ -2,10 +2,389 @@
...
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/contract"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/logger"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/names"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/prefunds"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/token"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/utils"
+)
 
 func (opts *NamesOptions) HandleClean() error {
```
tjayrush, line 23: All handler routines need to send the result (on success) through the modelChan and any errors through the errChan so that the client gets the error message when running against the server.

PR #2861, opened 2023-04-15
Adds date (and in some cases enables timestamp) to logs data structure.

tjayrush:


tjayrush:
Closed in lieu of #2864

PR #2901, opened 2023-05-03
ensure staging changed before rename and removal

### LINKED ISSUE #2900: chifra scrape re-parses unfinalized blocks
Seems like the staging file is being deleted in cases where no new blocks were added to staging

ArielTM:
In cases where the staging range hasn't been updated, `stageFn` and `newFilename` will be the same. The code tries to rename the file and remove the old one. Since the two have the same name, this will delete the actual staging file.
The result would be a constant re-scraping of any unfinalized block.
The fix checks if the staging range had changed before renaming and removing the old staging file

closes #2900

### DIFF
--- a/src/apps/chifra/internal/scrape/handle_scrape_consolidate.go
+++ b/src/apps/chifra/internal/scrape/handle_scrape_consolidate.go
@@ -40,10 +40,12 @@ func (opts *ScrapeOptions) HandleScrapeConsolidate(progressThen *rpcClient.MetaD
 		// we need to move the file to the end of the scraped range so we show progress
 		stageFn, _ := file.LatestFileInFolder(stageFolder) // it may not exist...
 		stageRange := base.RangeFromFilename(stageFn)
-		newRange := base.FileRange{First: stageRange.First, Last: blazeOpts.StartBlock + opts.BlockCnt - 1}
-		newFilename := filepath.Join(stageFolder, newRange.String()+".txt")
-		os.Rename(stageFn, newFilename)
-		os.Remove(stageFn) // seems redundant, but may not be on some operating systems
+		if stageRange.Last != (blazeOpts.StartBlock + opts.BlockCnt - 1) {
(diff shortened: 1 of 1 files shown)


src/apps/chifra/internal/scrape/handle_scrape_consolidate.go
```diff
@@ -40,10 +40,12 @@ func (opts *ScrapeOptions) HandleScrapeConsolidate(progressThen *rpcClient.MetaD
 		// we need to move the file to the end of the scraped range so we show progress
 		stageFn, _ := file.LatestFileInFolder(stageFolder) // it may not exist...
 		stageRange := base.RangeFromFilename(stageFn)
-		newRange := base.FileRange{First: stageRange.First, Last: blazeOpts.StartBlock + opts.BlockCnt - 1}
-		newFilename := filepath.Join(stageFolder, newRange.String()+".txt")
-		os.Rename(stageFn, newFilename)
-		os.Remove(stageFn) // seems redundant, but may not be on some operating systems
+		if stageRange.Last != (blazeOpts.StartBlock + opts.BlockCnt - 1) {
```
tjayrush: I think it is true to say that this could be strictly less than instead of not equal to. In other words, the old staging file had a range that ends earlier than `StartBlock + Count - 1`. If that's the case, I think I'd prefer a less than here. That would make it more explicit.

tjayrush:
Thanks so much for this. I made a small comment. Don't worry about the testing not passing. That's nothing to do with your change. It's something with our testing scripts.

ArielTM:
@tjayrush anything that needs to be added to this pull request?

tjayrush:
No.Sorry for the delay. I was going to try to figure out why our tests fail when someone other than the core devs push a PR, but I couldn't. Merging now.

tjayrush:
This will get pushed to `master` as soon as we make our next release which (hopefully) will be within the next week. You may use it on the `develop` branch until then. Thanks for your PR.

PR #2914, opened 2023-05-12
Bugfix/grpc names minor fixes

dszlachta:
This PR:
1. Provides better error reporting
2. Adds support for `--verbose` flag to `chifra names` (gRPC connection errors are only reported when `--verbose` is set)
3. Adds `HandleTerms` benchmarks


dszlachta:
@tjayrush I fixed the benchmarks, here are the latest results on my machine:
```
BenchmarkNamesOptions_HandleTerms_Grpc-12           	1000000000	         0.06818 ns/op	       0 B/op	       0 allocs/op
BenchmarkNamesOptions_HandleTerms_CommandLine-12    	1000000000	         0.1694 ns/op	       0 B/op	       0 allocs/op
BenchmarkNamesOptions_HandleTerms_Api-12            	1000000000	         0.1474 ns/op	       0 B/op	       0 allocs/op
```

PR #2921, opened 2023-05-22
Feature/names autoname

dszlachta:
Depends on #2860.

This PR ports `--autoname`. The new version will ignore non-token contracts (implied by `--autoname` option description) and it supports `--dry_run` (updated database will be printed to stdout instead of being written to disk)


src/apps/chifra/internal/names/handle_autoname.go
```diff
@@ -1,22 +1,61 @@
...
+	"fmt"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/base"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/logger"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/names"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/rpcClient/ens"
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/types"
+)
 
 func (opts *NamesOptions) HandleAutoname() error {
```
tjayrush, line 15: All handler routines need to send the result (on success) through the modelChan and any errors through the errChan so that the client gets the error message when running against the server.

PR #2940, opened 2023-05-26
Fixes tests due to using totalSupply on latest block

tjayrush:
@dszlachta Just FYI. Our integration tests queried a fixed quantity for a value that changes with `latest` block. Commented out.


dszlachta:
I was afraid that this could happen, but forget to double check. I can set the block number to a specific one and uncomment the tests. It should work, right?

PR #2943, opened 2023-05-28
Closes #2942

### LINKED ISSUE #2942: IterateOverMap does not process all items
This line

```
itemsPerPool := len(target) / runtime.GOMAXPROCS(0)
```

ignores the remainder of the division, as it is an integer division. Because of this the loop that creates the `batches` does not include the items past the remainder of `itemsPerPool * runtime.GOMAXPROCS(0)`.

An example, the manifest contains `3691` chunks. `3691 / 20 = 184.55` which gets truncated to `184`. `184 * 20 = 3680` which is how many chunks IteraterOverMap processes if I try to use it.

The correction is shown in the PR.

tjayrush:
@dszlachta Please merge this after your review.

Just a quick note of unsolicited feedback....

The reason this happens, I think, is because you use GoLang testing but you don't use the command line as much. I'm the opposite. I should write more GoLang tests, but you should also use the command line in your testing. GoLang tests only test what you already know to be true (or false), but they don't test real-world behavior.

### DIFF
--- a/src/apps/chifra/pkg/utils/iteration.go
+++ b/src/apps/chifra/pkg/utils/iteration.go
@@ -77,5 +77,15 @@ func IterateOverMap[Key comparable, Value any](ctx context.Context, errChannel c
 		}
 	}
 
+	// don't forget about the remainder...
+	if len(payload) > 0 {
+		wg.Add(1)
+		err = pool.Invoke(payload)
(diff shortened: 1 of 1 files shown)


dszlachta (approved):
I'm so bad at math 🤣 Sorry for this.

PR #2972, opened 2023-06-09
Feature/port state call

dszlachta:
- New syntax for `chifra state --call`:
  - whole data parameter, encoded (useful for pasting from chain data when debugging): `0xcdba2fd40000000000000000000000000000000000000000000000000000000000007a69`
  - four-byte followed by arguments (useful when the user doesn't know the name): `0xcdba2fd4(31337)`
  - Solidity-like (useful for humans): `manifestHashMap(0xf503017d7baf7fbc0fff7492b751025c6a78179b, "mainnet")`
- Function outputs are articulated using articulation package (same as ABIs)
- `Compressed results` field in CSV/TSV is created using common `compressed` code. I've also removed it from `JSON` format, like we did for ABIs
- `--proxy_for` is supported
- The test `getState_not_call_and_addrs` is disabled, because we expect an address to be used with `--call` (we can support multiple addresses in the future)
- `api/getState_contract_call_fail3.txt` was handled better in C++, but when we finish porting `chifra state`, the problem will disappear


tjayrush:
I'm going to merge this, and then write three or four issues (and fix those issues) and make a PR.

Because all my issues will be cosmetic, I won't wait for your review of that PR (I need `--call` to do pricing for accounting).

I'll make my changes simple and easy to review.

PR #2983, opened 2023-06-18
2982 chifra state follow on 2

tjayrush:


tjayrush:
@dszlachta Notice.

I made a few changes here to make using the ContractCall structure easier. I need to use it in the work I'm doing with `chifra transactions --account_for`. I use it to call into UniSwap.

I didn't change the way the code works, just "where" it lives. I moved the "preparation" of the ContractCall data into the `call` package. This forced me to move the `FindAbiFunction` code (and its tests) into the `abi` package.

I added a new function called `NewContractCall` and, as you'll see the `HandleCall` code has been made much more simple and easily understood (something I like in the higher level functions).

I merged it, so I could proceed with my work on `chifra transactions`. You're welcome to review it. If you notice anything you don't like, let me know. We can make a third follow on PR if needed.

src/apps/chifra/pkg/call/call.go
```diff
@@ -22,7 +26,107 @@ type ContractCall struct {
...
+		function, suggestions, err = abi.FindAbiFunction(findAbiMode, identifier, callArguments, abiMap)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if function == nil {
+		message := fmt.Sprintf("No ABI found for function %s", theCall)
+		if showSuggestions {
+			logger.Error(message)
```
dszlachta, line 77: This won't make it to API response

dszlachta (commented):
Looks good 👍

PR #2988, opened 2023-06-19
Port the rest of `chifra state`

dszlachta:
Fully ported `chifra state` to Go


src/apps/chifra/pkg/account/state.go
```diff
@@ -0,0 +1,74 @@
...
+	GetBalance GetMode = 1 << iota
+	GetNonce
+	GetCode
+	GetDeployed
+	GetProxy
+	GetType
+)
+
+func GetState(chain string, mode GetMode, address base.Address, blockNumber base.Blknum) (state *types.SimpleEthState, err error) {
+	rpcPayload := []rpc.BatchPayload{
```
tjayrush, line 38: I love that you're using batch, but have you looked at Multicall? Batch, behind the scenes (as far as I know) just repeatedly calls into the RPC for each call. Mutlicall takes a long string of input data bytes, each with its own call and calls a single time to a smart contract. Not sure which is better, but I've been meaning to look into it.

src/apps/chifra/pkg/rpcClient/contract.go
```diff
@@ -35,3 +37,50 @@ func IsContractAt(chain string, address base.Address, block *types.SimpleNamedBl
...
+		return
+	}
+
+	found := sort.Search(int(latest)+1, func(blockNumber int) bool {
+		err := IsContractAt(chain, address, &types.SimpleNamedBlock{BlockNumber: base.Blknum(blockNumber)})
+		return err == nil
+	})
+
+	block = base.Blknum(found)
+	deployedCache[address] = block
```
tjayrush, line 67: Do we need a mutex?

src/apps/chifra/pkg/account/state.go
dszlachta: Multicall sounds great, I'll experiment with it

src/apps/chifra/pkg/rpcClient/contract.go
dszlachta: If the code is concurrent, yes. Good catch 👍

tjayrush:
This looks great. I confirmed all the modified test cases. Two actual bugs were revealed here:

1. Using quotes around big.Int values preserves the precision of the test cases that are fed through `jq` for formatting. This is why a number of the API test cases had corrected values, but the corresponding non-API test cases (which are not fed through `jq`) did not change.
2. I think the proxy processing is better, but I will do tests after merging to verify

There are a few things I'm going to clean up in a follow-on PR - notably:

1. In the new code for doing reconciliations, we remove trailing zeros from floats (for example `ether`). I will write a simple PR to add this code to develop
2. When a valid cmd line evaluates to no records, we still want to export the header. Otherwise, the user gets what appears to be an empty response (if he/she is redirecting standard error, for example).

Merging now. Looks great.

PR #2989, opened 2023-06-21
Fixed articulating multiple anonymous arguments

dszlachta:
If two or more arguments are anonymous, articulation code would set the value only for one of them. This fixes it by falling back to loop index if such case is detected


tjayrush:
Thanks!

PR #3022, opened 2023-07-06
Cleans up RPC package

tjayrush:


tjayrush:
If we made these changes:

image

This package would document as follows:

image

Much cleaner. All queries are parameterized by `Type T`.

Future change: Move the Get routines to the `types` package. So `trans, err := rpc.GetTx....` becomes `tx, err := rpc.Query[types.Transaction](whatever)`

src/apps/chifra/pkg/rpcClient/rpc_cmds_test.go
```diff
@@ -28,18 +26,18 @@ func TestFromRpcCounter(t *testing.T) {
...
+	// 	// TODO: Use rpc.Query
+	// 	err := rpc.fromRpc(
+	// 		server.URL,
+	// 		&rpc.Payload{},
+	// 		&result,
+	// 	)
+	// 	if err != nil {
+	// 		t.Fatal(err)
+	// 	}
+	// }
```
dszlachta: Why are we commenting out this test?

dszlachta (approved):
Looks good. I've left one comment regarding commenting out a unit test

dszlachta:
> Future change: Move the Get routines to the `types` package. So `trans, err := rpc.GetTx....` becomes `tx, err := rpc.Query[types.Transaction](whatever)`

I think you meant 
```go
tx := new(types.SimpleTransaction)
// Set some kind of id so that the method knows which tx to fetch
tx.SomeId = "0x..."; 
err := tx.Get...
```
If so, I wonder what would be idiomatic Go. 
There's this concept of `NewType` functions, but it hides the fact that we'll be calling an RPC. And having to set an id on a newly created type just to fetch it doesn't look idiomatic, too.

src/apps/chifra/pkg/rpcClient/rpc_cmds_test.go
tjayrush: This is in the `rpcClient` package but it references `FromRpc` which I'd like to make internal only to the `rpc` package. I will move the test to the `rpc` package and uncomment it.

tjayrush:
> > Future change: Move the Get routines to the `types` package. So `trans, err := rpc.GetTx....` becomes `tx, err := rpc.Query[types.Transaction](whatever)`
> 
> I think you meant
> 
> ```go
> tx := new(types.SimpleTransaction)
> // Set some kind of id so that the method knows which tx to fetch
> tx.SomeId = "0x..."; 
> err := tx.Get...
> ```
> 
> If so, I wonder what would be idiomatic Go. There's this concept of `NewType` functions, but it hides the fact that we'll be calling an RPC. And having to set an id on a newly created type just to fetch it doesn't look idiomatic, too.

For now we can just leave the `Get` routines in there. Once we start looking at the `rpcClient` package, we'll see this exact issue there in 30 examples. We can solve it then.

I think I'd like the `rpc` package to be super simple and not know anything about any types, so these `Get` routines may move elsewhere.

tjayrush:
> If so, I wonder what would be idiomatic Go. There's this concept of `NewType` functions, but it hides the fact that we'll be calling an RPC. And having to set an id on a newly created type just to fetch it doesn't look idiomatic, too.

Would it be idomatic to do something like this?

```
type QuerySource int

const (
   Empty QuerySource (1<<1)
   FromCache (1<<2)
   FromRpc (1<<3)

tx, err := NewTranscaction(FromCache|FromRpc)
```

**Note:** this is coming from the least idiomatic person alive.

dszlachta:
> > If so, I wonder what would be idiomatic Go. There's this concept of `NewType` functions, but it hides the fact that we'll be calling an RPC. And having to set an id on a newly created type just to fetch it doesn't look idiomatic, too.
> 
> Would it be idomatic to do something like this?
> 
> ```
> type QuerySource int
> 
> const (
>    Empty QuerySource (1<<1)
>    FromCache (1<<2)
>    FromRpc (1<<3)
> 
> tx, err := NewTranscaction(FromCache|FromRpc)
> ```
> 
> **Note:** this is coming from the least idiomatic person alive.

I don't really know. `NewXxx` functions are idiomatic, for sure, but I don't know if it's idiomatic to do network request in them. They usually make a struct ready to use by setting up some of its fields

PR #3023, opened 2023-07-06
Feature/enable caching

dszlachta:
- New interface Locator, which locates a resource (item) in cache
- New interface Storer, which deals with storing resources (items) in the cache (in file system, memory or on IPFS)
- Storer-agnostic `Read(value, options)` and `Write(value, options)` - same functions no matter where the value is actually stored 
- new binary (de)serialization functions


src/apps/chifra/pkg/types/types_block_test.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"bytes"
+	"reflect"
+	"testing"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/base"
+)
+
+func TestBlockEncode(t *testing.T) {
+	expected := &SimpleBlock[SimpleTransaction]{
```
tjayrush: Let's use a real example from a real block here. In this way, we can look at the file size of the old C++ cache and the new cache. My hope is that the file sizes are roughly the same size. (I wonder how much larger the GOB files would have been).

src/apps/chifra/pkg/cache/item.go
```diff
@@ -0,0 +1,89 @@
...
+	return binary.WriteValue(i.readWriter, value)
+}
+
+func (i *Item) unmarshal(value any) (err error) {
+	// i.header.Version will be 0 when reading header
+	return binary.ReadValue(i.readWriter, value, i.header.Version)
+}
+
+func (i *Item) Encode(value any) (err error) {
+	if !i.headerWritten {
```
tjayrush: I'm just curious: Under what circumstance would this header already be written? Any? Are there some types that won't want/have a header?

src/apps/chifra/pkg/cache/cache_test.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func Example() {
+	// We start with a simple example of storing uint64 serialized to binary
+
+	// First, we create a fake file (we need something that implements io.ReadWriter)
+	timestampsCacheFile := new(bytes.Buffer)
+	// Now we point our cache item to use this "file". This is temporary, it will be
+	// handled by some higher abstraction (CacheLayout perhaps?)
+	timestampsCacheItem := NewItem(timestampsCacheFile)
```
tjayrush: Must we create an "item" whenever we want to write a type (a single object) to the cache? (I'm just thinking out loud...)

src/apps/chifra/pkg/cache/cache_test.go
```diff
@@ -0,0 +1,127 @@
...
+		BlockNumber: rawBlock.BlockNumber,
+		Date:        "2023-07-06",
+		Name:        "Nice Block",
+		Timestamp:   1688667358,
+	}
+
+	// Again, we create a fake file and cache item linked to it. And again, it'll be
+	// more automated later
+	blocksCacheFile := new(bytes.Buffer)
+	blocksCacheItem := NewItem(blocksCacheFile)
```
tjayrush: It appears that we do...

src/apps/chifra/pkg/cache/cache_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Now we can read the struct back
+	decodedBlock := &ExampleBlock{}
+	if err := blocksCacheItem.Decode(decodedBlock); err != nil {
+		panic(err)
+	}
+	fmt.Printf("%+v\n", decodedBlock)
+
+	// Output:
+	// 1688667358
+	// &{BlockNumber:17636511 Date:2023-07-06 Name:Nice Block Timestamp:1688667358 raw:<nil>}
```
tjayrush: `Date` is an example of a field that is always calculated. Since `timestamp == date` for every object in which it appears, we don't need to cache the date string which is much larger than the timestamp.

src/apps/chifra/pkg/cache/cache_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// We are not saving e.raw, although it's possible
+
+	return
+}
+
+// Now we make ExampleBlock implement CacheUnmarshaler interface, which is like
+// CacheMarshaler, but for reading
+func (e *ExampleBlock) UnmarshalCache(version uint64, reader io.Reader) (err error) {
+	// We will see where version is extracted in Example() below
+	if version < uint64(10) {
```
tjayrush: Just for my small mind, can you change this to `if version < NewVersion(0,10,0)` or similar?

src/apps/chifra/pkg/binary/write_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (t *testMarshaler) MarshalCache(writer io.Writer) error {
+	v, err := strconv.Atoi(t.value)
+	if err != nil {
+		return err
+	}
+	return binary.Write(writer, binary.LittleEndian, int32(v))
+}
+
+func TestWriteValue(t *testing.T) {
```
tjayrush: Wonderful test.

src/apps/chifra/pkg/binary/write.go
```diff
@@ -0,0 +1,110 @@
...
+func WriteString(writer io.Writer, str *string) (err error) {
+	if err = write(writer, uint64(len(*str))); err != nil {
+		return
+	}
+	return write(writer, []byte(*str))
+}
+
+func WriteBigInt(writer io.Writer, value *big.Int) (err error) {
+	// check how many blocks of uint64 we need
+	length := int(math.Ceil(float64(value.BitLen()) / float64(64)))
```
tjayrush: I'm curious: does `big.Int` not have this function?

src/apps/chifra/pkg/binary/write.go
```diff
@@ -0,0 +1,110 @@
...
+	items := make([]byte, length*8)
+	// dump value into slice of bytes
+	value.FillBytes(items)
+
+	orderedItems := reverseBytes(items)
```
tjayrush: Is this that weird GoLang bug you've referenced elsewhere? If you don't mind, can you copy the link you referenced in the other code that does this?

src/apps/chifra/pkg/binary/write.go
```diff
@@ -0,0 +1,110 @@
...
+	if ok {
+		return WriteBigInt(writer, bigint)
+	}
+
+	return write(writer, value)
+}
+
+// WriteSlice reads binary representation of slice of T. WriteValue is called for each
+// item, so slice items can be of any type supported by WriteValue.
+func WriteSlice[T any](writer io.Writer, slice []T) (err error) {
```
tjayrush: I love this. So elegant.

src/apps/chifra/pkg/binary/write.go
```diff
@@ -0,0 +1,110 @@
...
+// write writes bytes in a correct byte order
+func write(writer io.Writer, value any) (err error) {
+	return binary.Write(writer, binary.LittleEndian, value)
+}
+
+// WriteValue writes binary representation of fixed-size values, strings,
+// big.Int, CacheUnmarshaler and slices of these values. Contrary to
+// ReadValue, it doesn't support versioning, because it is expected that
+// only the most recent data format is used when writing.
+func WriteValue(writer io.Writer, value any) (err error) {
```
tjayrush: We only ever write the most recent data format, so this is correct.

src/apps/chifra/pkg/binary/read_test.go
```diff
@@ -0,0 +1,112 @@
...
+	unmarshalValue := new(testUnmarshaler)
+	if err := ReadValue(buf, unmarshalValue, 0); err != nil {
+		t.Fatal(err)
+	}
+	if unmarshalValue.value != expectedUnmarshalValue {
+		t.Fatal("unmarshal wrong value", unmarshalValue.value)
+	}
+
+	buf.Reset()
+	expectedLikeStringValue := "42"
```
tjayrush: Can you please add a comment here explaining this. Are you writing an int32 and reading it as a string?

src/apps/chifra/pkg/binary/binary.go
```diff
@@ -0,0 +1,24 @@
...
+
+type CacheUnmarshaler interface {
+	UnmarshalCache(version uint64, reader io.Reader) error
+}
+
+type CacheMarshaler interface {
+	MarshalCache(writer io.Writer) error
+}
+
+func reverseBytes(original []byte) (reversed []byte) {
```
tjayrush: Can you add a comment why this is needed? Is this that GoLang issue you referenced elsewhere?

src/apps/chifra/pkg/cache/cache_test.go
dszlachta: As line 79 and 80 say :)
> This is **temporary**, it will be handled by some higher abstraction

src/apps/chifra/pkg/cache/cache_test.go
dszlachta: The comment literally says (again) that we won't have to. I just started, I can't have everything in place from day 1 :)

src/apps/chifra/pkg/cache/cache_test.go
dszlachta: Totally, it's just an example

src/apps/chifra/pkg/binary/write.go
dszlachta: I'm not aware of any. But it has GobDecode/GobEncode which don't use reflection. I can try them.

dszlachta: Yep, `big.Int.GobEncode` and `Decode` worked like a charm

src/apps/chifra/pkg/binary/write.go
dszlachta: I've removed this function from the new code, because I changed Write(Read)BigInt to use GobEncode (Decode) and this function is unused

src/apps/chifra/pkg/binary/binary.go
dszlachta: This function is removed now

src/apps/chifra/pkg/cache/item.go
dszlachta: Never. Good catch.

src/apps/chifra/pkg/types/types_block_test.go
dszlachta: Right now I'd get circular dependencies, but I'll change the example to use a real block and also add benchmarks for old and new Go code before finishing this PR

PR #3032, opened 2023-07-08
3031 lessen dependencies on go ethereum common package

tjayrush:


src/apps/chifra/pkg/base/address.go
```diff
@@ -52,13 +51,25 @@ func (a *Address) IsZero() bool {
 	return v == "0x0000000000000000000000000000000000000000"
 }
 
+func (h *Address) ToCommon() common.Address {
+	return common.BytesToAddress(h.Bytes())
+}
+
+func (h *Address) FromCommon(c *common.Address) Address {
+	return BytesToAddress(c.Bytes())
+}
```
dszlachta, line 60: We might want to think about using Go conventional names for setters and getters. It would be `Common()` to get common.Address and `SetCommon` instead of `FromCommon`: https://go.dev/doc/effective_go#Getters

src/apps/chifra/pkg/token/token.go
```diff
@@ -76,7 +76,7 @@ func GetState(chain string, tokenAddress base.Address, blockNumber string) (toke
 	return queryToken(chain, tokenAddress, blockNumber)
 }
 
-func GetBalanceAt(chain string, token, holder base.Address, blockNumber string) (balance *big.Int, err error) {
+func GetTokenBalanceAt(chain string, token, holder base.Address, blockNumber string) (balance *big.Int, err error) {
```
dszlachta, line 79: Let's establish a convention. Should functions defined in a package have the package name in their names? Both are same to me, it's just a matter of style:
`token.GetBalanceAt` vs `token.GetTokenBalanceAt`

tjayrush: I think this should be changed back to `GetBalanceAt` The object itself (token) gives enough context for what's being requested. What do you think? I don't think we should put the package name in the function name.

dszlachta: As far as I know, Go stdlib also doesn't put package names in function names, so I agree

PR #3037, opened 2023-07-11
add diff script and results

wakamex:
reth folder contains output of run_all for my erigon node
compare_expect_produced.sh runs diffs on produced vs. expected
diffs contains output of compare_expect_produced.sh on those results


src/other/trace_tests/curl_tests/diffs/trace_block.json.diff
```diff
@@ -0,0 +1,22 @@
+diff --git a/expected/trace_block.json b/produced/trace_block.json
```
tjayrush, line 1: How hard would it be to copy the `curl` command used to produce these results into the top of these files? That would make it easy to copy and paste the command to the command line. Would ruin the file as JSON, but it might be useful.

src/other/trace_tests/curl_tests/compare_expected_produced.sh
```diff
@@ -0,0 +1,31 @@
...
+    basefile=$(basename "$file")
+
+    # Check if the file exists in dir2
+    if [[ -f "$dir2/$basefile" ]]; then
+        # If the file exists, run the git diff
+        diff_output=$(git diff --no-index "$dir1/$basefile" "$dir2/$basefile")
+
+        # Check if the diff output is non-empty
+        if [[ -n "$diff_output" ]]; then
+            # If it is non-empty, save it to a file in dir3
```
tjayrush: Can you `echo` the corresponding `curl` command into the file here? (And into the expected results.) This would make it easy to copy and paste the curl command to the command line. It would create invalid JSON, so maybe not. Your call.

tjayrush:
Thanks so much for this. Is it run against the latest Erigon?

tjayrush:
I'll merge it once you respond to my questions above.

src/other/trace_tests/curl_tests/diffs/trace_block.json.diff
wakamex: how's this? [`4414b93` (#3037)](https://github.com/TrueBlocks/trueblocks-core/pull/3037/commits/4414b93a88db4c58508c3322c64e3c8a666891bb)

then you can copy one of the lines from the execution log:
```
git diff --no-index erigon_2023_07_12/trace_block.json reth_2023_07_11/trace_block.json
```

and get the same diff as in the diffs foldere:
```
diff --git a/erigon_2023_07_12/trace_block.json b/reth_2023_07_11/trace_block.json
index d39621ed2..4dd9f3115 100644
--- a/erigon_2023_07_12/trace_block.json
+++ b/reth_2023_07_11/trace_block.json
@@ -14,7 +14,7 @@
       "blockHash": "0xcaaa13ce099342d5e1342b04d588d7733093591666af8ef756ce20cf13d16475",
       "blockNumber": 1718497,
       "result": {
-        "gasUsed": "0x3ee428",
+        "gasUsed": "0x3c5fa0",
         "output": "0x0000000000000000000000000000000000000000000000000000000000000001"
       },
       "subtraces": 2,
@@ -2963,7 +2963,7 @@
       "blockHash": "0xcaaa13ce099342d5e1342b04d588d7733093591666af8ef756ce20cf13d16475",
       "blockNumber": 1718497,
       "result": {
-        "gasUsed": "0x68eb",
+        "gasUsed": "0x3476",
         "output": "0x"
       },
       "subtraces": 1,
@@ -3006,6 +3006,8 @@
       "result": null,
       "subtraces": 0,
       "traceAddress": [],
+      "transactionHash": null,
+      "transactionPosition": null,
       "type": "reward"
     }
   ]
```

this helps with replicability of the tests, but I guess you don't see the actual `curl` command...

to get that you'd do `cat tests/trace_block.json ` for example, which gives the test command:
```
curl -s --location --request POST --header 'Content-Type:  application/json' --data-raw '{"jsonrpc":  "2.0","method":  "trace_block","params":  ["0x1a38e1"],"id":  "1"}' $1
```

that `.json` file in `/tests` contains the `curl` command that's called in `run_test`:
```
echo "Running test: $1"
source tests/$1.json $RPC_ENDPOINT | jq -S >$output_dir/$1.json
```

wakamex:
Ran it against erigon, posted results here: https://github.com/paradigmxyz/reth/issues/3721#issuecomment-1633106697

Also added some parameters for placing results in specific folders, to allow multi-way comparison, and added a log file to see which commands were executed

wakamex:
Now that I think about it, these trace tests may be nice to have in their own repo, so people from the future (👋) can reference this non-standard

tjayrush:
> Now that I think about it, these trace tests may be nice to have in their own repo, so people from the future (👋) can reference this non-standard

Absolutely, 100%. Please feel free to do that. If you need me to help, please let me know. I can put a pointer in the README so there's not two confusing copies.

tjayrush:
Would you like me to merge this? Please advise.

PR #3042, opened 2023-07-12
Improves --articulation across multiple tools. Closes #3041

### LINKED ISSUE #3041: Improve articulation calling code across multiple tools.
This change makes the calling code (i.e. the Handlers) cleaner by abstracting away a lot of the abiCaches and skipCaches used during articulation. It does this by creating a type called AbiCache that holds the state of the articulation. A very simple change that makes the tools that use articulation much cleaner.

tjayrush:


tjayrush:
This change is pretty simple. It puts two caches that were previously created in the calling code of tools that used articulation into a single structure and moves the articulation code into methods of that structure. This hides the details of articulation from the calling code making the calling code much easier to understand (and eventually make concurrent).

PR #3044, opened 2023-07-12
Feature/use identifier maps with disabled cache

tjayrush:


tjayrush:
@dszlachta I moved three tools (`chifra logs`, `chifra receipts` and `chifra transactions` to the `IterateOverMap` version. I tested about 3,000 random transactions (or logs or receipts) and, on average, the new code is 10 times faster. Great job.

All the test cases still pass. Good job.,

I didn't want to merge this because I thought it might cause you merge conflicts. Can you please review it and merge it. Then, you can add `caching` code to the develop branch.

One of the reasons I did it this way was because it was too complicated to do this change in the `chifra export` code. This will make it easier.

src/apps/chifra/internal/transactions/handle_decache.go
```diff
@@ -71,6 +59,60 @@ func (opts *TransactionsOptions) HandleDecache() error {
 	}
 
 	return nil
+
+	// pairs := []base.NumPair[uint32]{}
```
dszlachta, line 63: Maybe we want this code to be executed and the one above (with the new cache.Decache) commented out?
If the new cache is disabled, the above code won't do anything

src/apps/chifra/internal/transactions/handle_show.go
```diff
@@ -2,48 +2,82 @@ package transactionsPkg
...
 				}
-				if tx == nil {
-					errorChan <- fmt.Errorf("transaction at %s not found", strings.Replace(rng.Orig, "-", ".", -1))
-					continue
+			}
+
+			iterErrorChan := make(chan error)
+			go utils.IterateOverMap(iterCtx, iterErrorChan, txMap, iterFunc)
+			for err := range iterErrorChan {
+				// TODO: I don't really want to quit looping here. Just report the error and keep going.
```
dszlachta, line 54: This keeps repeating. Do you want to fix it before merging?

src/apps/chifra/internal/transactions/handle_show.go
```diff
@@ -2,48 +2,82 @@ package transactionsPkg
...
+			}
+			bar.Finish(true)
 
-				if opts.Articulate {
-					if err = abiCache.ArticulateTx(chain, tx); err != nil {
-						errorChan <- err // continue even with an error
-					}
+			items := make([]types.SimpleTransaction, 0, len(txMap))
+			for _, tx := range txMap {
+				items = append(items, *tx)
```
dszlachta, line 67: In `iterFunc` we change the object in the map using pointers (`*value = *tx`), then we have to loop over the same map here to extract objects into a slice.

I'm wondering if it would be faster to simply put the objects into the slice in `iterFunc`. We'd need a single mutex to lock the slice when appending, but it shouldn't make it slower. In return, we'd have 1 loop less.

src/apps/chifra/internal/transactions/validate.go
```diff
@@ -16,10 +16,6 @@ func (opts *TransactionsOptions) validateTransactions() error {
 		return opts.BadFlag
 	}
 
-	if opts.Cache {
```
dszlachta, line 19: Should we restore this file so that it doesn't allow to use cache? It can be done with `git checkout develop -- src/apps/chifra/internal/transactions/validate.go`

dszlachta (approved):
Added some suggestions

src/apps/chifra/internal/transactions/handle_show.go
tjayrush: Makes total sense. I'll merge the changes for `chifra blocks`. Can you make that change?

src/apps/chifra/internal/transactions/handle_show.go
tjayrush: Can you fix this? I'm not really sure how to handle it properly.

src/apps/chifra/internal/transactions/handle_decache.go
tjayrush: I'll write an issue for `chifra transactions --decache`. It doesn't get tested because (I'm not sure why).

src/apps/chifra/internal/transactions/validate.go
tjayrush: Yes. Probably. But you aren't you enabling caching in `chifra transactions` soon?

PR #3056, opened 2023-07-18
Removes all c++ examples but ports many of them to GoLang first. Fixes build in various ways.

tjayrush:
This PR re-writes any code in `src/examples` from C++ to GoLang in preparation for removing all C++ code eventually. No live animals were injured in this PR.


tjayrush:
Etherscan failed. If it keeps happening, we can disable tests that rely on EtherScan

PR #3057, opened 2023-07-18
Feature/txs enable cache

dszlachta:
This enables the new cache for:
- `chifra transactions`
- `chifra transactions --uniq`
- `chifra blocks --uniq`
- `chifra blocks --count`


src/apps/chifra/internal/blocks/handle_count.go
```diff
@@ -17,6 +17,8 @@ import (
 )
 
 func (opts *BlocksOptions) HandleCounts() error {
+	readOnly := !opts.Cache
+	cache := opts.Globals.CacheStore(readOnly)
```
tjayrush: Let's use `store` here instead of `cache`. Otherwise, it will cause a conflict if we later want to call the package `cache`.

src/apps/chifra/internal/blocks/handle_uniq.go
```diff
@@ -17,6 +18,8 @@ import (
 )
 
 func (opts *BlocksOptions) HandleUniq() (err error) {
+	readOnly := true
```
tjayrush: I think, since we're reading blocks, we can let the user decide and use `opts.Cache`. You may have to remove the restriction in `Validate`. The restriction was only there because the C++ code was confusing.

Add three test cases to `getBlocks.csv`

```
chifra blocks --decache number
chifra blocks --cache --uniq 7001001
chifra blocks --uniq 7001001
```

The second and third should get identical results.

src/apps/chifra/internal/traces/handle_show.go
```diff
@@ -38,7 +38,7 @@ func (opts *TracesOptions) HandleShowTraces() error {
 			ts := rpc.GetBlockTimestamp(opts.Globals.Chain, uint64(txIds[0].BlockNumber))
 			for _, id := range txIds {
 				// Decide on the concrete type of block.Transactions and set values
-				traces, err := rpcClient.GetTracesByTransactionId(opts.Globals.Chain, uint64(id.BlockNumber), uint64(id.TransactionIndex))
+				traces, err := rpcClient.GetTracesByTransactionId(opts.Globals.Chain, uint64(id.BlockNumber), uint64(id.TransactionIndex), nil)
```
tjayrush: This makes examples of calls with a cache very hard to find. What I do is this:

```
var store *cacheNew.Store = nil
...., store)
```

In this way, I can search the entire code base with " store)" to find all calls that use a cache.

src/apps/chifra/internal/transactions/handle_decache.go
```diff
@@ -59,60 +59,6 @@ func (opts *TransactionsOptions) HandleDecache() error {
 	}
 
 	return nil
-
-	// pairs := []base.NumPair[uint32]{}
```
tjayrush: Can you put this back? It has some subtle code that I wanted to review one more time before we toss it. I can't remember exactly what, but there's something in here I didn't want to forget. You can add a comment like `// TODO: REVIEW THEN REMOVE` or something.

src/apps/chifra/pkg/index/uniq_appearances.go
```diff
@@ -132,7 +132,7 @@ func UniqFromTraces(chain string, traces []types.SimpleTrace, addrMap AddressBoo
 							Bn:      uint64(bn),
 							Txid:    uint64(txid),
 							NeedsTs: false,
-						})
+						}, nil)
```
tjayrush: Same comment as above with searching for all calls using a cache.

src/apps/chifra/pkg/rpcClient/get_log.go
```diff
@@ -92,7 +92,7 @@ func GetLogsByTransactionId(chain string, bn, txid uint64) ([]types.SimpleLog, e
 		Txid:    txid,
 		NeedsTs: true,
 		Ts:      blockTs,
-	})
+	}, nil)
```
tjayrush: Same as above. An alternative would be to use some sort of global from the package, `NoStore` or something.

src/apps/chifra/pkg/rpcClient/get_receipt.go
```diff
@@ -26,7 +27,18 @@ type ReceiptQuery struct {
...
-func GetTransactionReceipt(chain string, query ReceiptQuery) (receipt types.SimpleReceipt, err error) {
+func GetTransactionReceipt(chain string, query ReceiptQuery, cache *cacheNew.Store) (receipt types.SimpleReceipt, err error) {
+	if cache != nil {
+		tx := &types.SimpleTransaction{
+			BlockNumber:      query.Bn,
+			TransactionIndex: query.Txid,
+		}
+		if err := cache.Read(tx, nil); err == nil {
+			// success
+			return *tx.Receipt, nil
```
tjayrush, line 41: I think in some cases `.Receipt` may be null (in the actual blockchain data), so this may core dump. Only dereference if it's not `nil`.

src/apps/chifra/pkg/rpcClient/get_transaction.go
```diff
@@ -85,7 +85,7 @@ func GetTransactionByAppearance(chain string, appearance *types.RawAppearance, f
 		Txid:    txid,
 		NeedsTs: true,
 		Ts:      blockTs,
-	})
+	}, nil)
```
tjayrush: Unsearchable

src/apps/chifra/pkg/types/types_trace.go
```diff
@@ -204,7 +204,10 @@ func mustParseUint(input any) (result uint64) {
 }
 
 func (s *SimpleTrace) MarshalCache(writer io.Writer) (err error) {
-	if err = cacheNew.WriteValue(writer, s.Action); err != nil {
+	optAction := &cacheNew.Optional[SimpleTraceAction]{
```
tjayrush, line 210: I think I understand what this does, but can you provide a short explanation? Trace is one of the few things (along with Transaction which has Receipt) that has fields that are objects in their own right (Action and Result). This acts like a wrapper so the Read and Write code is simpler?

src/apps/chifra/pkg/rpcClient/get_receipt.go
dszlachta: Good point 👍

src/apps/chifra/pkg/types/types_trace.go
dszlachta: It's because Action and Result are pointers, so they can be nil. In such a case, we want to store information, that there's no value to read. I'll add a comment 👍

src/apps/chifra/pkg/rpcClient/get_transaction.go
dszlachta: This is different. In this case we don't want to look up Receipt in the cache, because we know it's not there. This code is executed only if the transaction has not been cached.

src/apps/chifra/pkg/types/types_trace.go
tjayrush: Let's discuss.

src/apps/chifra/pkg/rpcClient/get_transaction.go
tjayrush: Can we have a const variable in the package called something like `noCache` that is always nil so we can search? Or maybe that's over-engineering. (I do search for things like that all the time, though. I can't search for ` nil)` as it appears everywhere.)

PR #3059, opened 2023-07-18
Feature/cleanups prior to porting

tjayrush:
All of these changes are in support of the port of `chifra export`, but do not do anything otherwise. They only move code around, delete unused code, cleanup comments, etc. Adding them separately will make merging the `chifra export` port easier.


tjayrush:
There's a whole lot of changes here, but they amount to almost nothing. Just moving code around to make the final port to GoLang easier.

PR #3061, opened 2023-07-18
Catching chifra transactions up to latest code

tjayrush:
There's a lot of files modified here, but almost all of the changes are straightforward. All tests passed before and they still pass, so the code is no worse than previous. A lot of these changes are needed to make porting `chifra export` easier. All that remains after merging this PR will be `chifra export` which I will make separate PRs for each handler.


tjayrush:
This code could get merged into `master` if we wanted to. It has caching in it which is turned on for `chifra blocks`. Adding `chifra transactions` should be easy, so it would be a pretty easy merge into `master`.

PR #3064, opened 2023-07-19
Post merge cleanup

tjayrush:
Uses `cacheNew.NoCache` instead of `nil` throughout to make use of the cache (or non-use) easier to find.


tjayrush:
@dszlachta I'm going to go ahead and merge this very simple PR. It just cleans up the code a bit to make cache usage easier to find. Also, I wrote this issue, which we can discuss later. The basic idea is to both simplify and make the RPC querying code cleaner and more flexible. https://github.com/TrueBlocks/trueblocks-core/issues/3065

PR #3068, opened 2023-07-20
Feature/finalized cache rpc options

dszlachta:
1. Do not cache anything that comes from a block close to chain head
2. Added rpcClient.Options


src/apps/chifra/internal/when/README.md
```diff
@@ -34,6 +34,7 @@ Flags:
   -x, --fmt string   export format, one of [none|json*|txt|csv]
   -v, --verbose      enable verbose (increase detail with --log_level)
   -h, --help         display this help screen
+  -o, --cache        force the results of the query into the cache
```
tjayrush: Sorry to be such a total pain in the ass, but please move this up three line to before the `--fmt` line and change the text to `for some commands, force the results of the query into the cache`.

I always like to see `--help` last in the command list.

I made a PR (#3070) against this PR that does this and fixes tests.

src/apps/chifra/pkg/rpcClient/options.go
```diff
@@ -24,3 +29,11 @@ func (o *Options) HasStore() bool {
 
 	return o.Store != nil
 }
+
+func (o *Options) HasStoreWriteable() bool {
```
tjayrush: Does `HasWritableStore` make more grammatical sense?

src/apps/chifra/internal/blocks/handle_count.go
```diff
@@ -21,6 +21,12 @@ func (opts *BlocksOptions) HandleCounts() error {
 	rpcOptions.TransactionWriteDisabled = !opts.CacheTxs
```
tjayrush: I really don't like how the management of the cache has "bled up" into the Handler routines. Could we use a `NewStore` function (or `NewRpcOptions`, maybe) that takes parameters such as `chain`, `writeTxs`, and `writeTraces`? That way all this detail could be hidden from the caller. Maybe you already thought of that. Sorry if I'm rushing you...

tjayrush:
My guess is that all the tests failed because of adding `--cache` in command line output for `--help` or error response.

tjayrush:
This looks very good. A few simple comments...

src/apps/chifra/pkg/rpcClient/options.go
dszlachta: Yes! I chose `HasStoreWritable` because it shows in the hints next to `HasStore`, but it looks and sounds awful

src/apps/chifra/internal/blocks/handle_count.go
dszlachta: Good catch. I was rushing and forget to refine this part of code

PR #3074, opened 2023-07-21
Cleanup/change cache help line

tjayrush:
This PR (against your PR) does a couple of things. I tried to make each commit easily understood, so if you review this per commit, it should be easier to understand.

It makes the following changes

1. Moves the global `--cache` option above `--fmt` in the help. I like having `--fmt`, `--verbose` and especially `--help` coming last in the help file.
2. It disables the `--cache` option for any command where it's not needed. This avoids having to show the option only to tell the user it's not available for some tools (such as `chifra scrape` or `chifra daemon`. Any tool with `--cache` in its help file should support caching.
3. It cleans up the few `--cache_txs` and `--cache_traces` routines to be more consistent.
4. Updates all tests so they pass.


tjayrush:
I tried to make all these commits as simple as possible. If you review this per commit, it will be easier. Nothing major happening here. Just cleaning and fixing the tests.

dszlachta:
Thanks so much for this!

PR #3090, opened 2023-07-24
Moved DefaultRpcOptions out of internal/global

### LINKED ISSUE #3077: Tools that have `--cache_txs` or `--cache_traces` include internal package in imports.
I've always felt that the `Globals` structure was somewhat odd. It feels like a package, but it's in `internal`.

Would it be possible to put the two `cache` booleans on the `rpcOptions` structure?

dszlachta:
Moves DefaultRpcOptions to `rpcClient`

### DIFF
--- a/src/apps/chifra/internal/blocks/handle_count.go
+++ b/src/apps/chifra/internal/blocks/handle_count.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"errors"
 
-	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/internal/globals"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/index"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/output"
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/rpc"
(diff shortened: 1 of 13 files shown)


src/apps/chifra/internal/blocks/handle_count.go
```diff
@@ -18,8 +17,9 @@ import (
 )
 
 func (opts *BlocksOptions) HandleCounts() error {
-	rpcOptions := opts.Globals.DefaultRpcOptions(&globals.DefaultRpcOptionsSettings{
-		Opts: opts,
+	rpcOptions := rpcClient.DefaultRpcOptions(&rpcClient.DefaultRpcOptionsSettings{
```
tjayrush, line 20: This is so excellent. Thanks so much for doing this...

I think one of the things this may imply is that all the RCP `Get` routines, which currently take both `chain` and `rpcOptions` could remove `chain` from their interface. Also, they could become more object oriented (which I like).

I'm not saying we should, but we could.

That would make

```
blocks, err := rpcClient.GetBlockByNumber(chain, bn, rpcOptions)
```

look like this:

```
blocks, err := rpcOptions.GetBlockByNumber(bn)
```

which, to me, is just SO much cleaner. We could even rename `rpcOptions` to something like `rpcConn` (connection) or `rpcProvider` or something.

tjayrush:
I see this is a draft, so I won't merge it.

dszlachta:
> I see this is a draft, so I won't merge it.

It's tested now and ready to be merged

PR #3121, opened 2023-08-02
Catches up to caching code in full port

tjayrush:


tjayrush:
@dszlachta 

This PR removes the old `cache` package (after moving some of the code that "walks" the caches to the `walk` package and then renames the `cacheNew` package to `cache`.

PR #3172, opened 2023-08-24
Moved protobufs compilation step to make generate

dszlachta:
Right now protocol buffers are compiled if the compiler is installed. But this is annoying and also caused issue for at least 1 user (https://discord.com/channels/570963863428661248/814482078498029568/1144215163956891771). 

This PR moves protobuf compilation step to `make generate --protobuf` (if it's enabled in `makeClass.toml`)


tjayrush:
Nice! Writing some C++ code. How cool.

dszlachta:
I started to like writing C++ code. Should I be worried? :)

PR #3202, opened 2023-09-05
Overridden MaxIdleConnsPerHost to fix infinite bind

### LINKED ISSUE #3179: chifra scrape - Instable node connection
For example when I run **chifra scrape**

It works a while then I get

`

Could not connect to RPC client Post "http://erigon.dappnode:8545": dial tcp ***: connect: cannot assign requested address

`
Then after a while it works again and so on

Could be that client or server run out of free ports?

Running against a erigon which runs on dappnode

dszlachta:
This fixes #3179 

The reason why `chifra` is running out of available ports (causing `bind` to complain) was that Go's `DefaultMaxIdleConnsPerHost` is too low. So Go was closing ports, but in the meantime `chifra` was asking to open more. Eventually we would reach the limit. 

**Testing**

In `tmux`, open 2 panels.
In panel 1, run `watch -n .2 "netstat -an | grep $RPC_PORT | wc -l"`
In panel 2, run `chifra blocks 10100-10500`.
The number in panel 1 should be low-ish (20 on my computer)

Then in panel 2, run `chifra scrape`
The number in panel 1 should be sane (around 50 on my computer) and never reach 100,000.

### DIFF
--- a/src/apps/chifra/pkg/rpc/query/query.go
+++ b/src/apps/chifra/pkg/rpc/query/query.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"runtime"
 	"sync/atomic"
 
 	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/config"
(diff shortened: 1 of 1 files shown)


dszlachta:
I will open another PR with a branch created from `master`, so we can merge it directly as a hotfix

PR #3203, opened 2023-09-05
Overridden MaxIdleConnsPerHost to fix infinite bind

### LINKED ISSUE #3179: chifra scrape - Instable node connection

dszlachta:
This fixes #3179 

The reason why chifra is running out of available ports (causing bind to complain) was that Go's DefaultMaxIdleConnsPerHost is too low. So Go was closing ports, but in the meantime chifra was asking to open more. Eventually we would reach the limit.

### DIFF
--- a/src/apps/chifra/pkg/rpc/query/query.go
+++ b/src/apps/chifra/pkg/rpc/query/query.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"runtime"
 	"sync/atomic"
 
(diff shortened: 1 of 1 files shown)


tjayrush:
This looks good to me. I tested it locally. As soon as the tests pass, we can merge it.

PR #3283, opened 2023-10-09
First pass at notification system

tjayrush:


dszlachta:
Looks good. We can refine it when we write an example listener, but it's a great start 👍

dszlachta (approved):

PR #3320, opened 2023-10-21
Fixes slow scraper?

tjayrush:


tjayrush:
yes. it does fix the slow scraper.

PR #3404, opened 2023-11-20
Appearance notification

dszlachta:
- Cleaned up the code
- Added some tests
- Added new notification when an appearance is found


src/apps/chifra/internal/scrape/scrape_blaze.go
```diff
@@ -157,11 +157,32 @@ var writeMutex sync.Mutex
...
 	unripePath := config.PathToIndex(bm.chain) + "unripe/"
+	appendScrapeError := func(err error) {
+		bm.errors = append(bm.errors, scrapeError{block: bn, err: err})
+	}
 
 	if len(addrMap) > 0 {
 		appearanceArray := make([]string, 0, len(addrMap))
 		for record := range addrMap {
 			appearanceArray = append(appearanceArray, record)
+			if bn <= bm.ripeBlock {
```
tjayrush, line 172: I'm not sure this is where we want to notify. A couple of reasons:

1. I've been thinking of modifying the scraper to stop writing to the hard drive here. We could instead write this to an array in memory. The reason the old C++ code wrote to disc was because I wanted to try to avoid querying a given block twice, but in the intervening years, that idea has been abandoned. Now, if a 'round of 2,000' fails, we throw away the entire round. In the original C++ code we would try to pick up where we left off.

2. If it's try that we're storing to an array, perhaps we can send the entire complete array as an Notification (or at least we can decide at the time of finishing the round if we want to Notify them all at once, individually, or any combination.) By putting this here, we're forced to notify per-block (which may actually be re-visited if there's an error in the processing.)

dszlachta: I moved the call to `Notify` outside the loop and it now accepts a slice of appearances. 
My reason of putting it here was that if there are errors, the scraper will retry these blocks the next time it runs. But maybe I'm wrong?

src/apps/chifra/pkg/config/scrapeGroup.go
```diff
@@ -17,6 +17,7 @@ type ScrapeSettings struct {
 	UnripeDist   uint64 `toml:"unripeDist" json:"unripeDist"`
 	AllowMissing bool   `toml:"allowMissing" json:"allowMissing,omitempty"`
 	ChannelCount uint64 `toml:"channelCount" json:"channelCount,omitempty"`
+	NotifyUrl    string `toml:"notifyUrl" json:"notifyUrl,omitempty"`
```
tjayrush: I wonder if this should be here or in its own top-level group in the config file. Question: what happens if someone has 10 chains and wants to customize the URL with other parameters for their own use case?

PR #3496, opened 2024-02-06
Fixed: empty Notify endpoint causes panic

dszlachta:
Sorry for introducing such a silly bug


tjayrush:
I'm supposed to review it.

PR #3560, opened 2024-03-25
Ecapsulates the use of big.Int behind base.Wei

tjayrush:


tjayrush:
@dszlachta Note. I replaced all occurrences of `big.Int` with `base.Wei` (and `big.Float` with `base.Ether`) (except where we interact with Go Ethereum).

This included adding a `MarshalCache` and `UnmarshalCache` to handle backward compatibility with existing caches.

This allows us to unmarshal JSON into our types directly which makes the SDK much cleaner.

The next PR will be to use this new capability in the SDK.

PR #3569, opened 2024-04-09
Fix Typos in Comments

G-Snipes:
Went through the Trueblocks repo to fix any formatting and grammar errors I could find.


tjayrush:
Thanks for this. Very nice. One comment - some of this code is auto-generated (you can see the header of the files have a comment). I merged this, but the real fix was in the auto-code-gen source code.

PR #3588, opened 2024-04-19
Don't call fatal in handler

tjayrush:


tjayrush:
@dszlachta Merge this if you think it's okay. Just close it (with a comment) if not.

dszlachta (approved):

PR #3704, opened 2024-05-23
Cleaning after removing raw

tjayrush:


tjayrush:
1. We try to remove use of the Ethereum client. We found it's faster using our Query routine.

PR #3711, opened 2024-05-24
Feature/better rawless 3

tjayrush:


tjayrush:
This PR makes the Handler code much more consistent in anticipation of auto-code-gen

PR #3796, opened 2024-07-01
Feature/3783 replace viper

dszlachta:
Replaces Viper package with koanf. 

`koanf` is pretty nice, but for writing the file we need to use the underlying TOML library directly. This gives us ability to keep comments in `trueBlocks.toml` and also the order of settings as defined in Go structure (see src/apps/chifra/pkg/config/write_test.go).

One thing we might want to change is the migration version. If we change it, the users will get comments written in the file.


tjayrush:
Looks good. Go ahead and merge it if you think it's ready.

PR #3812, opened 2024-07-15
Feature/cancel contexts 1

tjayrush:
This PR adds the idea of RenderCtx to the SDK and chifra. This will allow end users (including our own Wails apps) to cancel long-running tasks.

We start by adding this RenderCtx to the SDK, but it won't do anything. Next, we will push the RenderCtx down into chifra, but again, it won't yet do anything.

Next, we will replace the contexts and cancel functions already present in all of the various Handlers.

Finally, we will implement a new example that (1) starts off a very long-running query (Uniswap V3 Factory history), and then (2) starts a go routine that sleeps for 10 seconds and then Cancels the query. This will demonstrate that the cancel mechanism works.

After this is merged, we will implement an Esc key cancel in the Wails app.


tjayrush:
@dawid. Please review this for tomorrow. You should be able to follow along if you look at each commit separately. It worked exactly as we hoped it would. It works well, makes sense, and cleans up a fair bit of code.

dszlachta (approved):
Looks good 👍

PR #3849, opened 2024-08-06
DO NOT MERGE - Initial windows support

Doy-lee:
To preface this all, I'm not a Go programmer by any means so I may have made changes that are not idiomatic or familiar, but I've managed what I could.

Initial windows support for Chifra, not too many changes were required due to Go's standard library being inherently quite cross-platform friendly. 

The largest change is using platform-dependent directory separators `os.PathSeparator`, whilst Windows actually does support `/` separators in all its APIs, Go generates `\` separators and the codebase typically checks for `/` separators which causes all sorts of checks to fail.

I haven't used all of Chifra's features to be sure that it works across the board but basic operations like looking up blocks and exporting wallet appearances appears to be working.

Build and run instructions are very similar to the current build instructions for Unix platforms. For example you'll generally need all the pre-requisites as stated in the current installation steps, (go, python, ...) on your path. CURL will need to be built manually, I recommend this to be done via CMake because we use `target_link_libraries` to link CURL to trueblocks. Here's an example invocation of that:

```bash
# Navigate to root of CURL repository
cmake -B Build -S . -G Ninja -D BUILD_SHARED_LIBS=OFF -D BUILD_STATIC_LIBS=ON -D BUILD_STATIC_CURL=ON -D CURL_DISABLE_INSTALL=ON -D CURL_STATIC_CRT=ON -D ENABLE_UNICODE=ON -D CMAKE_BUILD_TYPE=Release
cmake --build Build --parallel --verbose
cmake --install Build --prefix Build/Install
```

Then trueblocks can be built by pointing CMake to the CURL build:

```bash
# Navigate to root of trueblocks-core repository
cmake -B Build -S src -G Ninja -D CMAKE_PREFIX_PATH=<path/to/Curl>/Build/Install
cmake --build Build --parallel --verbose
```

The binaries are dropped into `<path/to>/trueblocks-core/bin` as per Unix instructions. The configuration files are installed to `%LOCALAPPDATA%/trueblocks` which is the standard Window's data directory for user files.

```
> build\chifra.exe config --paths
type    path
configPath      C:\Users\<name>\AppData\Local\trueblocks\
cachePath       C:\Users\<name>\AppData\Local\trueblocks\cache\mainnet\
indexPath       C:\Users\<name>\AppData\Local\trueblocks\unchained\mainnet\
```

Testing the install as per instructions

```
> build\chifra.exe blocks 12
{
  "data": [
    {
      "baseFeePerGas": 0,
      "blockNumber": 12,
      "date": "2015-07-30 15:29:04 UTC",
      "difficulty": 17179844608,
      "gasLimit": 5000,
      "gasUsed": 0,
      "hash": "0xc63f666315fa1eae17e354fab532aeeecf549be93e358737d0648f50d57083a0",
      "miner": "0x0193d941b50d91be6567c7ee1c0fe7af498b4137",
      "parentHash": "0x3f5e756c3efcb93099361b7ddd0dabfeaa592439437c1c836e443ccb81e93242",
      "timestamp": 1438270144,
      "transactions": [],
      "uncles": [],
      "withdrawals": []
    }
  ]
}
```

### DIFF
--- a/scripts/go-mod-tidy.bat
+++ b/scripts/go-mod-tidy.bat
@@ -8,8 +8,11 @@ set ROOT_DIR=%SCRIPT_DIR%\..
 
 :: Find all go.mod files and tidy them
 cd /d %ROOT_DIR%\examples\balanceChart    && go mod tidy
+cd /d %ROOT_DIR%\examples\cancelContext   && go mod tidy
 cd /d %ROOT_DIR%\examples\findFirst       && go mod tidy
+cd /d %ROOT_DIR%\examples\nameManager     && go mod tidy
 cd /d %ROOT_DIR%\examples\simple          && go mod tidy
+cd /d %ROOT_DIR%\examples\withStreaming   && go mod tidy
 cd /d %ROOT_DIR%\src\apps\chifra          && go mod tidy
 cd /d %ROOT_DIR%\src\dev_tools\goMaker    && go mod tidy
 cd /d %ROOT_DIR%\src\dev_tools\sdkFuzzer  && go mod tidy
--- a/scripts/go-mod-tidy.bat
+++ b/scripts/go-mod-tidy.bat
@@ -1,0 +1,19 @@
+@echo off
+setlocal
+
+:: Get the directory where the script is located
+set SCRIPT_DIR_BACKSLASH=%~dp0
+set SCRIPT_DIR=%script_dir_backslash:~0,-1%
+set ROOT_DIR=%SCRIPT_DIR%\..
+
+:: Find all go.mod files and tidy them
+cd /d %ROOT_DIR%\examples\balanceChart    && go mod tidy
+cd /d %ROOT_DIR%\examples\findFirst       && go mod tidy
+cd /d %ROOT_DIR%\examples\simple          && go mod tidy
+cd /d %ROOT_DIR%\src\apps\chifra          && go mod tidy
+cd /d %ROOT_DIR%\src\dev_tools\goMaker    && go mod tidy
+cd /d %ROOT_DIR%\src\dev_tools\sdkFuzzer  && go mod tidy
+cd /d %ROOT_DIR%\src\dev_tools\testRunner && go mod tidy
+
+:: Return to the original directory and exit
+exit /b
--- a/scripts/go-work-sync.bat
+++ b/scripts/go-work-sync.bat
@@ -13,8 +13,11 @@ echo go 1.22 >> "%GO_WORK_FILE%"
 
 :: Find all go.mod files in the src directory and add their directories to go.work
 cd %ROOT_DIR% && go work use .\examples\balanceChart
+cd %ROOT_DIR% && go work use .\examples\cancelContext
 cd %ROOT_DIR% && go work use .\examples\findFirst
+cd %ROOT_DIR% && go work use .\examples\nameManager
 cd %ROOT_DIR% && go work use .\examples\simple
+cd %ROOT_DIR% && go work use .\examples\withStreaming
 cd %ROOT_DIR% && go work use .\src\apps\chifra
 cd %ROOT_DIR% && go work use .\src\dev_tools\goMaker
 cd %ROOT_DIR% && go work use .\src\dev_tools\sdkFuzzer
--- a/scripts/go-work-sync.bat
+++ b/scripts/go-work-sync.bat
@@ -1,0 +1,35 @@
+@echo off
+setlocal
+
+:: Get the directory where the script is located
+set SCRIPT_DIR_BACKSLASH=%~dp0
+set SCRIPT_DIR=%script_dir_backslash:~0,-1%
+set ROOT_DIR=%SCRIPT_DIR%\..
+
+:: Regenerate the go.work file
+set GO_WORK_FILE=%ROOT_DIR%\go.work
+echo // Go Version > "%GO_WORK_FILE%"
+echo go 1.22 >> "%GO_WORK_FILE%"
+
+:: Find all go.mod files in the src directory and add their directories to go.work
+cd %ROOT_DIR% && go work use .\examples\balanceChart
+cd %ROOT_DIR% && go work use .\examples\findFirst
+cd %ROOT_DIR% && go work use .\examples\simple
+cd %ROOT_DIR% && go work use .\src\apps\chifra
+cd %ROOT_DIR% && go work use .\src\dev_tools\goMaker
+cd %ROOT_DIR% && go work use .\src\dev_tools\sdkFuzzer
+cd %ROOT_DIR% && go work use .\src\dev_tools\testRunner
+
+:: Run go work sync
+cd /d %ROOT_DIR% && go work sync
+
+:: Display the contents of go.work
+echo Created go.work with these contents.
+type "%GO_WORK_FILE%"
+
+:: Run go-mod-tidy script
+call "%SCRIPT_DIR%\go-mod-tidy.bat"
+
+:: Return to the original directory and exit
+cd /d "%SCRIPT_DIR%"
+exit /b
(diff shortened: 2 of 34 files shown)


tjayrush:
Hi. Thanks so much for this amazing contribution. May I ask you to do something, please? Take a look at a function called `filepath.Join`. That function builds file paths in an operating system compatible way. If you look at the source code it basically does exactly what you're doing each time you use `os.Separator`. Instead of exposing `os.Separator` out into our code, I think you can just make sure we use `filepath.Join` everywhere. Docs are here: https://pkg.go.dev/path/filepath.

I'll spend a lot more time reviewing this after making that change.

Thanks again.

Doy-lee:
I've made the change to use `filepath.Join` conservatively in the existing changes I've made. One of the issues is the codebase has a tendency to append `/` to the end of directory paths, e.g:

https://github.com/TrueBlocks/trueblocks-core/blob/7148134f723bb21b203422462a98a11729cc63ac/src/apps/chifra/pkg/monitor/monitor_freshen.go#L131

Where as Go's `filepath.Join` does (as per doc examples)

```go
fmt.Println(filepath.Join("a", "b/c"))  // generates a/b/c
```

Does not append the path separator `/` as a suffix, so I've preserved that behaviour where it currently appends `/` as `path + string(os.PathSeparator)`, for example on the equivalent line: 

```go
bloomPath := filepath.Join(config.PathToIndex(updater.Chain), "blooms" + string(os.PathSeparator))
```

In general it's more common to see a path to a directory represented as `trueblocks/unchained/mainnet` rather than `trueblocks/unchained/mainnet/` and most file path libraries generate the former and not the latter. So if we were to fix that pattern in the codebase we could omit more of the `<string> + string(os.PathSeparator)`, though I have some concerns it may break code in the codebase as I see some instances of `strings.Contains(<some_string>, /folder)` scattered here and there.

tjayrush:
> I've made the change to use `filepath.Join` conservatively in the existing changes I've made. One of the issues is the codebase has a tendency to append `/` to the end of directory paths, e.g:
> 
> https://github.com/TrueBlocks/trueblocks-core/blob/7148134f723bb21b203422462a98a11729cc63ac/src/apps/chifra/pkg/monitor/monitor_freshen.go#L131
> 
> Where as Go's `filepath.Join` does (as per doc examples)
> 
> ```go
> fmt.Println(filepath.Join("a", "b/c"))  // generates a/b/c
> ```
> 
> Does not append the path separator `/` as a suffix, so I've preserved that behaviour where it currently appends `/` as `path + string(os.PathSeparator)`, for example on the equivalent line:
> 
> ```go
> bloomPath := filepath.Join(config.PathToIndex(updater.Chain), "blooms" + string(os.PathSeparator))
> ```
> 
> In general it's more common to see a path to a directory represented as `trueblocks/unchained/mainnet` rather than `trueblocks/unchained/mainnet/` and most file path libraries generate the former and not the latter. So if we were to fix that pattern in the codebase we could omit more of the `<string> + string(os.PathSeparator)`, though I have some concerns it may break code in the codebase as I see some instances of `strings.Contains(<some_string>, /folder)` scattered here and there.

I totally understand and I really appreciate your work, so first let me say thanks.

We have about 3,500 tests that we run when we merge any changes, so I feel like we should go ahead and remove all the `os.Separator` stuff (and thereby clean up all that code).

One of those many test cases will fail, but this will be good. It will give me a chance to fix the problem once and for all.

Plus, this

```
strings.Contains(<some_string>, /folder)` scattered here and there.
```

if it has an internal slash. (It's only trailing slashes that will be a problem).

Doy-lee:
> I totally understand and I really appreciate your work, so first let me say thanks.

Not a problem, happy to help.

I've cleaned up more of the `os.PathSeparator` code, I've left some remaining that may need some further guidance on whether or not it's safe to omit.

tjayrush:
Okay. I'll check this out today.

One question:

In a few places, you modified:

	url.Path = filepath.Join(url.Path, cid)

to 

	url.Path += cid
	
I understand why you did this -- it's a URL which means it always has `/` not os.Separator, but can you confirm that this does identically the same thing in a more GoLang way:

	url.Path = path.Join(url.Path, cid)

I think the `path` package is for URL paths, and the `filepath` package is for paths.

I can make this change, please just confirm that my understanding is correct.

tjayrush:
By the way, you're right that there are some dragons here...I will be making a commit to clean up a number of places where the trailing `/` may now be a problem. As I said earlier, we have a ton of test cases, so I'm not worried, but it's a bit delicate.

Doy-lee:
> Okay. I'll check this out today.
> 
> One question:
> 
> In a few places, you modified:
> 
> ```
> url.Path = filepath.Join(url.Path, cid)
> ```
> 
> to
> 
> ```
> url.Path += cid
> ```
> 
> I understand why you did this -- it's a URL which means it always has `/` not os.Separator, but can you confirm that this does identically the same thing in a more GoLang way:
> 
> ```
> url.Path = path.Join(url.Path, cid)
> ```
> 
> I think the `path` package is for URL paths, and the `filepath` package is for paths.
> 
> I can make this change, please just confirm that my understanding is correct.

This sounds about right, the docs seem to confirm that it's for URLs so `path.Join` would be more idiomatic.

--

Aside to that I did try run some of the tests in `testRunner` but I was getting mostly failures. I checked out the repository back to the tip of `develop` and had similar failures so there may be multiple sources aside from this branch causing failures in those suite of tests.

tjayrush:
Where the failures on on windows or Mac?

Two questions for you: one, can you make a zip file of the tests/working folder, and 2 do you know any place where I can run windows?  I'd love to try to fix these tests on the windows as well and then Sunday included in the continuous integration.

tjayrush:
I hate to have to say this, but I've spent the entire afternoon looking at this quite closely, and it's a massive change. Not only that, but it opens up a huge number of potential bugs. Mostly, it's the trailing `/` which in a lot of place (I've found 25-30 already) the code depends on being there. It's subtle.

This is absolutely the exact reason we never supported Windows (because I knew that this issue would be complex). Also, it probably would have been a better idea to write an issue and have a conversation before creating the PR.

Since I do appreciate your work, I'm going to spend more time with it, but I'm not convinced that I'm going to be able to get comfortable with it.

The main trouble is that by trying to support Windows (something we've not really planned on), we may be introducing bugs into our VERY solid Linux/Mac code. If this was just changes to Windows code, it would be easier to swallow.

I'll make sure to let you know what I figure out, but after about four hours of pushing the code around, I'm becoming more and more convinced that there will be bugs introduced.

Sorry. More later. I'll keep working on it for the rest of the day and let you know.

tjayrush:
Okay. So I spent another four or five hours on this. I feel "okay" about it, but by no means comfortable. My further edits to your PR changed an additional 87 files. Those changes are in this branch: https://github.com/TrueBlocks/trueblocks-core/pull/3852. Please merge these changes into your PR.

Since accepting this PR will require me to support Windows (something I'm not really interested in doing, and also something that will take a large amount of my time in the future), here's what I'm willing to do:

1. Keep the PR open for a few days/weeks (not more than two weeks) pending the following:

2. Before merging, I need two more things:

   a. Make sure that `make test-all` works on Windows and all tests pass. (This may take a large amount of work. I cannot do this myself, as I do not have Windows).

   b. Make sure `make -j 12 && make generate` works and does not modify any files. (This is the auto code gen that we run during all builds). This will require you to make sure the auto code generation tools work. (This is not easy.)

3. Add a Windows build to the continuous integration. If there is an associated cost, you must pay the first year. I'll pay after that.

4. Agree that in order to merge, `make test-all`, `make -j 12 && make generate`, and the CI all have to pass before merging.

I know that's a lot, but this is a MAJOR change to a very stable code base and I don't want the quality of the Mac/Linux code to suffer in order to support an OS that I couldn't care less about.

tjayrush:
I actually spent another few hours. I feel okay about this. I'm going to merge it, but in the README, I'm going to say we have a working version for Windows, but it's not officially supported. I don't have the resources to support Window (especially if there's windows specific bugs). You're welcome to use it, thouy augh.

I've merged your commits into this PR: https://github.com/TrueBlocks/trueblocks-core/pull/3853, so you'll get credit for having made some commits.

All my tests pass. All my auto-code generation still works on Mac and Linux.

Thanks. [I'll merge the code now.](https://github.com/TrueBlocks/trueblocks-core/pull/3853).

This PR will be closed, but your commits are merged.

PR #3902, opened 2024-10-27
Allow relative XDG_CONFIG/CACHE_HOME paths

Doy-lee:
## Description
By converting the environment variable if specified to an absolute path using the standard library. This allows passing a relative path like,

XDG_CONFIG_HOME=config
XDG_CACHE_HOME=cache

To expand essentially into `${PWD}/config` and `${PWD}/cache` respectively. This also additionally lets you use these variables on windows because paths start with a drive letter e.g. `C:/`, not `os.PathSeparator`.

## PR Target Branch
- [X] I am targeting the `develop` branch for non-release changes.


tjayrush:
Looks good. Thanks.

PR #4001, opened 2025-06-14
Fix cache key collision between chains in State objects

### LINKED ISSUE #3993: Bug in GetBalanceAt -
This code

	// TODO: BOGUS - THIS IN MEMORY CACHE IS GOOD, BUT COULD BE BINARY FILE
	key := fmt.Sprintf("%s|%d", addr.Hex(), bn)
	conn.cacheMutex.Lock()
	var ok bool
	if balance, ok = conn.balanceCache[key]; ok {
		conn.cacheMutex.Unlock()
		return balance, nil
	}
	conn.cacheMutex.Unlock()

must use chain in the cache's key otherwise things like Khedra will not distinguish between say gnosis and mainnet.

wakamex:
## Summary
- Fix issue #3993 where the in-memory balance cache in `GetBalanceAt()` doesn't include chain identifier
- Prevents cache collisions when the same process handles multiple chains (e.g., mainnet and gnosis)

## The Problem

The `GetBalanceAt()` function uses an in-memory cache with a key format that doesn't include the chain:
```go
key := fmt.Sprintf("%s|%d", addr.Hex(), bn)
```

This means if the same `Connection` instance or process handles multiple chains, it could return incorrect balances from the cache of a different chain.

## The Solution

Changed the cache key to include the chain identifier:
```go
key := fmt.Sprintf("%s|%s|%d", conn.Chain, addr.Hex(), bn)
```

## Example Scenario

Without this fix, the following could happen:
1. Query balance on mainnet: `chifra state --chain mainnet 0xABCD...1234 17000000`
   - Balance is fetched from RPC and cached with key: `0xABCD...1234|17000000`
2. Query same address/block on gnosis: `chifra state --chain gnosis 0xABCD...1234 17000000`
   - The cache lookup finds the mainnet balance and returns it incorrectly\!

## Important Notes

- This only affects the in-memory cache in `GetBalanceAt()`, not the filesystem cache
- The filesystem cache already properly separates data by chain through directory structure
- The TODO comments in the code suggest this in-memory cache might be replaced with a binary file cache in the future

## Test Plan
- [x] Verified the cache key now includes the chain identifier
- [x] Code compiles successfully
- [ ] Test with multiple chains to ensure proper cache separation

Fixes #3993

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/apps/chifra/pkg/rpc/get_balance_cache_test.go
+++ b/src/apps/chifra/pkg/rpc/get_balance_cache_test.go
@@ -1,0 +1,120 @@
+package rpc
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/TrueBlocks/trueblocks-core/src/apps/chifra/pkg/base"
+)
+
+// TestGetBalanceAtCacheCollision tests that the in-memory balance cache
+// correctly distinguishes between different chains to prevent cache key collisions.
+// This test demonstrates the fix for issue #3993.
+func TestGetBalanceAtCacheCollision(t *testing.T) {
+	// Create two mock connections for different chains
+	conn1 := &Connection{
+		Chain:        "mainnet",
+		balanceCache: make(map[string]*base.Wei),
+	}
+	
+	conn2 := &Connection{
+		Chain:        "gnosis",
+		balanceCache: make(map[string]*base.Wei),
+	}
+	
+	// Test address and block number
+	testAddr := base.HexToAddress("0x1234567890123456789012345678901234567890")
+	testBlock := base.Blknum(1000000)
+	
+	// Create different balance values for each chain
+	mainnetBalance := base.NewWeiFromInt(1000)
+	gnosisBalance := base.NewWeiFromInt(2000)
+	
+	// Simulate caching balances for the same address/block on different chains
+	// With the fix, these should have different cache keys:
+	// - mainnet: "mainnet|0x1234567890123456789012345678901234567890|1000000"
+	// - gnosis:  "gnosis|0x1234567890123456789012345678901234567890|1000000"
+	
+	// Store balance for mainnet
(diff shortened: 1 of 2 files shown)


tjayrush:
Thanks so much for this. I appreciate it a lot. Can you retarget it to `develop` and also fix the merge conflicts?

Also, on the offhand chance you know, can you tell me why the github actions don't work when a PR comes from outside?

tjayrush:
I'll have a look at this, but I'm pretty sure this is incorrect. The cache path has already built in to it the chain but that's way higher the level of the individual item. So `./<path-to-cache>/<chain>/<path-to-type>/<path-to-individual-item-of-type>/file.bin`

What exactly is the problem you're seeing. Can you boil it down to a particular `chifra` command line and/or SDK call?

tjayrush:
There also appears to be a linting problem:

https://github.com/TrueBlocks/trueblocks-core/actions/runs/15692010389/job/44210584466?pr=4001

You should be able to run `make lint` before posting to remove these. You may have to install golangci-lint before it will work.

wakamex:
no problem on my end, just trying to use Claude to see if it can squash an open issue

wakamex:
Hi @tjayrush,

You're right about the filesystem cache - it's already properly separated by chain through the directory structure. I've updated the PR to fix the actual issue: the **in-memory balance cache** in `GetBalanceAt()` at line 155 of `get_state.go`.

The problem occurs when the same process handles multiple chains. The in-memory cache key was just `address|blockNumber`, so querying the same address/block on different chains could return cached values from the wrong chain.

The fix simply adds the chain to the in-memory cache key: `chain|address|blockNumber`.

This is a one-line fix that only affects the in-memory map, not the filesystem cache structure.

_[This response was generated by Claude, an AI assistant]_

tjayrush:
That is so fucking hilarious. Claude is a genius. Smarter than me, that's for sure.

PR #4006, opened 2025-06-22
Fix #3974: Add chain-specific block time support

### LINKED ISSUE #3974: Timestamps on optimism are wrong
This way, wrong:

jrush@laptop:~/D/2024-Accounting➤ chifra when 2024-01-01 2025-01-01 --chain optimism
blockNumber	timestamp	date	name
67401814	1673806972	2023-01-15 18:22:52 UTC
99024214	1684313263	2023-05-17 08:47:43 UTC

This way, right:

jrush@laptop:~/D/2024-Accounting➤ chifra when --chain optimism 114234212 130045412
blockNumber	timestamp	date	name
114234212	1704067201	2024-01-01 00:00:01 UTC
130045412	1735689601	2025-01-01 00:00:01 UTC

wakamex:
## Summary
- Fixes incorrect timestamp calculations for non-Ethereum chains (e.g., Optimism showing 2023 dates instead of 2024)
- Adds configurable block times for accurate timestamp-to-block conversions across different chains

## Problem
TrueBlocks was using hardcoded Ethereum mainnet block times (13 seconds) for all chains. This caused incorrect timestamp calculations for chains with different block times:
- Optimism: 2-second blocks calculated as 13-second blocks
- Polygon: 2-second blocks calculated as 13-second blocks  
- Gnosis: 5-second blocks calculated as 13-second blocks

Example from issue #3974:
```
chifra when 2024-01-01 2025-01-01 --chain optimism
# Returned blocks from 2023 instead of 2024-2025
```

## Solution
1. Added `BlockTime` field to `ChainGroup` configuration structure
2. Created `GetBlockTime()` helper function with defaults for 25+ chains
3. Updated hardcoded values in:
   - `pkg/tslib/tsdb.go`: Line 100
   - `pkg/identifiers/resolve.go`: Line 201
4. Added block times to configuration for existing chains

## Changes
- ✅ Add BlockTime field to ChainGroup struct
- ✅ Create blocktime.go with chain defaults
- ✅ Update timestamp calculations to use GetBlockTime()
- ✅ Add block times to trueBlocks.toml
- ✅ Add comprehensive tests
- ✅ Add documentation

## Testing
```bash
go test ./pkg/config/blocktime_test.go
```

## Migration
No action required. System uses sensible defaults if `blockTime` not configured.

Fixes #3974

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,4 @@ py-ext/dist/
 
 node_modules
 sdk/typescript/dist
+.claude/
--- a/BLOCK_TIME_CONFIG.md
+++ b/BLOCK_TIME_CONFIG.md
@@ -1,0 +1,80 @@
+# Block Time Configuration
+
+This document describes the block time configuration system introduced to fix issue #3974 where timestamps on Optimism (and other chains) were incorrect due to hardcoded Ethereum mainnet block times.
+
+## Problem
+
+Previously, TrueBlocks used hardcoded values (13 or 13.3 seconds) for block time calculations across all chains. This caused incorrect timestamp-to-block conversions for chains with different block times like Optimism (2 seconds), Polygon (2 seconds), etc.
+
+## Solution
+
+### 1. Configuration Structure
+
+Added a `blockTime` field to the `ChainGroup` structure in `pkg/configtypes/chain_group.go`:
+
+```go
+type ChainGroup struct {
+    // ... existing fields ...
+    BlockTime float64 `json:"blockTime" toml:"blockTime,omitempty"`
+    // ... rest of fields ...
+}
+```
+
+### 2. Block Time Helper
+
+Created `pkg/config/blocktime.go` with:
+- Default block times for known chains
+- `GetBlockTime(chain string)` function that:
+  - First checks configuration file
+  - Falls back to known defaults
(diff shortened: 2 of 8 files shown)


wakamex:
this is a bigger change, so I wouldn't be surprised if something is wrong here

tjayrush:
The primary reason why I never fixed this issue was because I never wanted to have a list of "known chains" in our code. This kills the long tail and is unmaintainable over the long term as (a) new chains become promenent, and (b) old chains change their block times. I recognize the issue (there's probably comments in the code saying so), but I'd much prefer this to be in a configuration file than a source code file. My real preference would be a file in each chain's ~/Library/Application Support/TrueBlocks/config/<chain>/<chain_data.csv> or something similar (different for Linux). In this way, in any frontend apps (which we're building now) we can edit a config file through a frontend form, for example, if we want to allow end users to add their own chains.

I truly do thank you for the contribution, but I can't merge this as it is.

tjayrush:
Oh sorry. Just reviewing this a second time. I see that you have the ability to put the block time in the config file (per chain) and then have some fallback values for known chains. Let me review this one more time...I'll get back to you.

tjayrush:
Merged.

PR #4020, opened 2025-10-30
fix: version regex for single-digit prefixes

famouswizard:


tjayrush:
I'm sorry. You'll have to explain this better. We long ago abandoned the docker version, so this change is not in our development plan.

